PR #103, opened 2019-09-14
Check for null

ChristofferGreen:
When trying to create a hello world triangle I ran into a crash with ppGlobalExtensionNames being null. Maybe I set things up wrong? If not, this code fixed it.

### DIFF
--- a/Graphics/GraphicsEngineVulkan/src/VulkanUtilities/VulkanInstance.cpp
+++ b/Graphics/GraphicsEngineVulkan/src/VulkanUtilities/VulkanInstance.cpp
@@ -110,8 +110,11 @@ namespace VulkanUtilities
 #endif
         };
 
-        for (uint32_t ext = 0; ext < GlobalExtensionCount; ++ext)
-            GlobalExtensions.push_back(ppGlobalExtensionNames[ext]);
+        if (ppGlobalExtensionNames != nullptr)
+        {
+            for (uint32_t ext = 0; ext < GlobalExtensionCount; ++ext)
+                GlobalExtensions.push_back(ppGlobalExtensionNames[ext]);
+        }
 
         for(const auto* ExtName : GlobalExtensions)
         {


DiligentGraphics:
What crash do you see with hello world example? `ppGlobalExtensionNames` comes from user-provided `EngineVkCreateInfo::ppGlobalExtensionNames` and is only used when `EngineVkCreateInfo::GlobalExtensionCount` is also non-zero. None of the examples use user-provided extensions.
This change will not hurt though

ChristofferGreen:
The crash was in the .push back (meaning, inside stl code). I'm guessing it's undefined behaviour due to array from nullptr.

I am mostly just copy pasting from the 00 win32 tutorial triangle example so currently I don't really know what I am doing, but the code I have is this:

```
  SwapChainDesc SCDesc;
  SCDesc.SamplesCount = 1;

#if VULKAN_SUPPORTED
  EngineVkCreateInfo EngVkAttribs;
  EngVkAttribs.EnableValidation = true; // debug
  auto* pFactoryVk = GetEngineFactoryVk();
  pFactoryVk->CreateDeviceAndContextsVk(EngVkAttribs, &This.device, &This.immediateContext);
  if (!This.swapChain && This.window->handle() != nullptr)
    pFactoryVk->CreateSwapChainVk(This.device, This.immediateContext, SCDesc, This.window->handle(), &This.swapChain);
  else {
    println_r("Error! Diligent::Diligent Vulkan setup failed.");
    exit(1);
  }
  println_b("Finished Vulkan setup.");
#else
  println_r("Error! Vulkan not supported by Diligent.");
#endif
```

DiligentGraphics:
That seems to be OK. I don't know why you were getting this error though as you don't specify any number of user-defined global extensions

ChristofferGreen:
pFactoryVk->CreateDeviceAndContextsVk(EngVkAttribs, ...
  VulkanUtilities::VulkanInstance::Create(..., EngineCI.ppGlobalExtensionNames
    new VulkanInstance(..., ppGlobalExtensionNames
       ppGlobalExtensionNames[ext]

So, if EngVkAttribs has a null ppGlobalExtensionNames from the start it will end up used as an array which is a null ptr within new VulkanInstance. No check for EngineVkCreateInfo::GlobalExtensionCount is done.

DiligentGraphics:
It will not end up used unless you also set `GlobalExtensionCount` to a non-zero (default) value.

ChristofferGreen:
Ah, sorry, I see what you mean now, wasn't paying close enough attention.

Did some printf debugging 

In my code
`  std::cout << "extensions: " << EngVkAttribs.GlobalExtensionCount << std::endl;;
  std::cout << "addess before: " << &EngVkAttribs.GlobalExtensionCount << std::endl;;
  pFactoryVk->CreateDeviceAndContextsVk(EngVkAttribs, &This.device, &This.immediateContext);
`

And  at the start of EngineFactoryVkImpl::CreateDeviceAndContextsVk
`    std::cout << "addess after: " << &_EngineCI.GlobalExtensionCount << std::endl;;
    std::cout << "2extensions: " << _EngineCI.GlobalExtensionCount << std::endl;;
    std::cout << "3extensions: " << EngineCI.GlobalExtensionCount << std::endl;;
`

This was the output:
`   extensions: 0
addess before: 0xc8fd1c
addess after: 0xc8fd18
2extensions: 1
3extensions: 1
`

So i clearly have some compiler problems, maybe linker related. I'll look into it. Thanks for the help

DiligentGraphics:
Hmm, that is weird. Are you building with MinGW on windows?

ChristofferGreen:
Yes, mingw that I got from chocolatey (package manager for windows apps). From print debugging it looks like the compiler has rearranged the memory location of the member variables in my code compared to diligent. I mean that the EngVkAttribs object is passed in via reference fine, it's memory location is the same, but the member .GlobalExtensionCount has a different address.
Very strange, looking at the way gcc is called the compile options don't differ in any way I could see that would cause that. I build diligent by adding it via externalproject (cmake auto download/build).

Maybe that compiler is just buggy.

Is there a recommended mingw distribution for compiling diligent? I tried visual studio and it worked great (meaning, compiles and this bug does not happen) but a few other mingws I found online all have compiler problems when compiling the windows specific parts of diligent.

DiligentGraphics:
There are all sorts of unexplainable issues when building with MinGW. I spent several days trying to fix this [specific one](https://github.com/DiligentGraphics/DiligentEngine/issues/31), but gave up eventually. I use Visual Studio for development, and can't recommend any MinGW distribution.
Usually such issues happen when build system does not detect the change properly and does not rebuild one of the object files, which keeps using old offsets. So the first thing I would try is to delete all intermediate files and do a clean rebuild.
Also, did you try to build and run any of the samples? They should work with MinGW.

ChristofferGreen:
Found the reason. Diligent (and some of the libraries it uses)  were compiled in c++11 mode while my code was compiled in c++14 mode. Yesterday I thought that could impossibly cause any problems since it would be such a normal use case that surely gcc would warn if there were incompatibilities. Well, I should learn that gcc cares not for usability. So today I recompiled my code in c++11 mode (I have very little code so only a few changes needed) and the addresses are all correct. So likely I will need to run Diligent as a dll if I want to compile with c++14, thats ok I guess.

DiligentGraphics:
That makes some sense, though I wouldn't expect struct size / member offsets to differ between 11 and 14. Likely there is something else going on that manifests itself as an issue with offsets.
All 3rd party libraries that Diligent uses are build in c++11 mode. I think that DLL would be the easiest way. You can build them in Visual Studio and then use in your project.

PR #107, opened 2019-10-25
Fix iOS platform detection

dolphineye:
It turns out that `${CMAKE_SYSTEM_NAME}` is should now `iOS` instead of `Darwin`. I'm not sure of the reason of this change, but this might be related to the latest iOS 12 updates or something specific in our project.

Anyway, at this point, I added a condition check for `iOS` to fix the `Unsupported platform` error we get , but I left the initial checks to avoid introducing breaking changes.

If you feel you can come up with a better idea, let me know ;)


DiligentGraphics:
I can't reproduce the problem on the system with the following configuration:
macOS Version: 10.15 Catalina
XCode Version 11.1 (11A1027)
Cmake: 3.15.4

`CMAKE_SYSTEM_NAME` [by default is set to](https://cmake.org/cmake/help/v3.15/variable/CMAKE_SYSTEM_NAME.html#system-name-for-host-builds) the value of `CMAKE_HOST_SYSTEM_NAME`, which can only [have the following values](https://cmake.org/cmake/help/v3.15/variable/CMAKE_HOST_SYSTEM_NAME.html#cmake-host-system-name): `Linux`, `Windows` and `Darwin`.

Does your CMake system sets the value of `CMAKE_SYSTEM_NAME`? I guess that correct fix might be to use `CMAKE_HOST_SYSTEM_NAME` instead of `CMAKE_SYSTEM_NAME`.

DiligentGraphics:
OK, so it turned out that CMake 3.15 finally added official support for iOS and I believe you generated the solution by specifying `-DCMAKE_SYSTEM_NAME=iOS` in the command line, not as described [here](https://github.com/DiligentGraphics/DiligentEngine#ios).
With your change the projects seem to be work OK, but third-party projects required to build Vulkan backend complain on unsupported platform. They still can be built using the tool-chain setup method.

dolphineye:
> OK, so it turned out that CMake 3.15 finally added official support for iOS and I believe you generated the solution by specifying `-DCMAKE_SYSTEM_NAME=iOS` in the command line, not as described [here](https://github.com/DiligentGraphics/DiligentEngine#ios).
> With your change the projects seem to be working OK, but third-party projects required to build Vulkan backend complain about unsupported platform. They still can be built using the tool-chain setup method.

Indeed, we specify `-DCMAKE_SYSTEM_NAME=iOS` in our project generation script. Thanks for the feedback.

DiligentGraphics:
> Indeed, we specify `-DCMAKE_SYSTEM_NAME=iOS` in our project generation script. Thanks for the feedback.

This will work now too, if you don't need to build Vulkan back-end.

dolphineye:
> > Indeed, we specify `-DCMAKE_SYSTEM_NAME=iOS` in our project generation script.
> 
> This will work now too, if you don't need to build Vulkan back-end.

If we'd want to, what would we need to change to make this work? Is this something that should be changed in Diligent Engine, or rather on our side?

DiligentGraphics:
You can use the [old approach](https://github.com/DiligentGraphics/DiligentEngine#ios) with manual toolchain specification. There are some issues in 3rd party projects when you use the new way: they need to update CMake files to detect iOS platform properly. This should not be a big problem as with the toolchain specification all third-party projects work fine on iOS already.

dolphineye:
> You can use the [old approach](https://github.com/DiligentGraphics/DiligentEngine#ios) with manual toolchain specification. There are some issues in 3rd party projects when you use the new way: they need to update CMake files to detect iOS platform properly. This should not be a big problem as with the toolchain specification all third-party projects work fine on iOS already.

Thanks for all these details ;)

DiligentGraphics:
https://github.com/KhronosGroup/SPIRV-Tools/issues/2994

DiligentGraphics:
https://github.com/KhronosGroup/SPIRV-Tools/issues/2994 is now fixed, so `-DCMAKE_SYSTEM_NAME=iOS` should work for Vulkan on iOS as well

dolphineye:
\o/

PR #117, opened 2020-01-22
Added native window handle wrapper for supported platforms

jayrulez:
Use NativeHandle in tests

There are a few things that I'm still considering.

1. The terminology for iOS is view rather than window but I kept it as WindowHandle rather than ViewHandle for consistency.

2. This abstraction can be used in the engine creation descriptor struct for OGL. However, since I am not able to test all platforms I am not comfortable making this change. 

I believe I need to forward declare HWND in Win32NativeWindow.h to use HWND instead of void* for the type. Including Windows.h there is problematic.

### DIFF
--- a/Graphics/GraphicsEngine/interface/GraphicsTypes.h
+++ b/Graphics/GraphicsEngine/interface/GraphicsTypes.h
@@ -35,6 +35,7 @@
 #include "../../../Primitives/interface/BasicTypes.h"
 #include "../../../Primitives/interface/DebugOutput.h"
 #include "../../../Primitives/interface/FlagEnum.h"
+#include "../../../Platforms/interface/NativeWindow.h"
 #include "APIInfo.h"
 
 /// Graphics engine namespace
@@ -1318,21 +1319,11 @@ struct EngineCreateInfo
 };
 typedef struct EngineCreateInfo EngineCreateInfo;
 
-
 /// Attributes of the OpenGL-based engine implementation
 struct EngineGLCreateInfo DILIGENT_DERIVE(EngineCreateInfo)
 
-    /// Native window handle
-
-    /// * On Win32 platform, this is a window handle (HWND)
-    /// * On Android platform, this is a pointer to the native window (ANativeWindow*)
-    /// * On Linux, this is the native window (Window)
-    void* pNativeWndHandle DEFAULT_INITIALIZER(nullptr);
-
-#if PLATFORM_LINUX
-    /// For linux platform only, this is the pointer to the display
-    void* pDisplay         DEFAULT_INITIALIZER(nullptr);
-#endif
+	/// Native window wrapper
+	NativeWindow Window;
 };
 typedef struct EngineGLCreateInfo EngineGLCreateInfo;
 
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ void InitializeDiligentEngine(HWND NativeWindowHandle)
 #endif
         auto* pFactoryOpenGL = GetEngineFactoryOpenGL();
         EngineGLCreateInfo EngineCI;
-        EngineCI.pNativeWndHandle = NativeWindowHandle;
+        EngineCI.Window.hWnd = NativeWindowHandle;
         pFactoryOpenGL->CreateDeviceAndSwapChainGL(
             EngineCI, &m_pDevice, &m_pImmediateContext, SCDesc, &m_pSwapChain);
     }
@@ -217,7 +217,7 @@ On Android, you can only create OpenGLES device. The following code snippet show
 ```cpp
 auto* pFactoryOpenGL = GetEngineFactoryOpenGL();
 EngineGLCreateInfo EngineCI;
-EngineCI.pNativeWndHandle = NativeWindowHandle;
+EngineCI.NativeWindow.pAWindow = NativeWindowHandle;
 pFactoryOpenGL->CreateDeviceAndSwapChainGL(
     EngineCI, &m_pDevice, &m_pContext, SCDesc, &m_pSwapChain);
 IRenderDeviceGLES *pRenderDeviceOpenGLES;
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -1,5 +1,6 @@
 ## Current Progress
 
+* Added `NativeWindow` wrapper and replaced `pNativeWndHandle` and `pDisplay` members with it in `EngineGLCreateInfo` (API Version 240053)
 * Added C Interface (API Version 240052)
 
 ## v2.4.d
--- a/Tests/DiligentCoreAPITest/include/TestingEnvironment.hpp
+++ b/Tests/DiligentCoreAPITest/include/TestingEnvironment.hpp
@@ -34,6 +34,7 @@
 #include "RefCntAutoPtr.hpp"
 #include "SwapChain.h"
 #include "GraphicsTypesOutputInserters.hpp"
+#include "NativeWindow.h"
 
 #include "gtest/gtest.h"
 
@@ -46,16 +47,8 @@ namespace Testing
 class TestingEnvironment : public ::testing::Environment
 {
 public:
-    struct NativeWindow
-    {
-        void* NativeWindowHandle = nullptr;
-#if PLATFORM_LINUX
-        void* Display = nullptr;
-#endif
-    };
-
-
     TestingEnvironment(RENDER_DEVICE_TYPE deviceType, ADAPTER_TYPE AdapterType, const SwapChainDesc& SCDesc);
+
     ~TestingEnvironment() override;
 
     // Override this to define how to set up the environment.
--- a/Graphics/GraphicsEngineOpenGL/src/GLContextAndroid.cpp
+++ b/Graphics/GraphicsEngineOpenGL/src/GLContextAndroid.cpp
@@ -237,7 +237,7 @@ GLContext::GLContext(const EngineGLCreateInfo& InitAttribs, DeviceCaps& deviceCa
     major_version_(0),
     minor_version_(0)
 {
-    auto* NativeWindow = reinterpret_cast<ANativeWindow*>(InitAttribs.pNativeWndHandle);
+    auto* NativeWindow = reinterpret_cast<ANativeWindow*>(InitAttribs.Window.pAWindow);
     Init(NativeWindow);
 
     FillDeviceCaps(deviceCaps);
--- a/Graphics/GraphicsEngineOpenGL/src/GLContextIOS.mm
+++ b/Graphics/GraphicsEngineOpenGL/src/GLContextIOS.mm
@@ -46,7 +46,7 @@ namespace Diligent
         //Or better yet, use the GL3 way to get the version number
         glGetIntegerv( GL_MAJOR_VERSION, &MajorVersion );
         glGetIntegerv( GL_MINOR_VERSION, &MinorVersion );
-        LOG_INFO_MESSAGE(Info.pNativeWndHandle != nullptr ? "Initialized OpenGLES " : "Attached to OpenGLES ", MajorVersion, '.', MinorVersion, " context (", GLVersionString, ", ", GLRenderer, ')');
+        LOG_INFO_MESSAGE(Info.Window.pNSView != nullptr ? "Initialized OpenGLES " : "Attached to OpenGLES ", MajorVersion, '.', MinorVersion, " context (", GLVersionString, ", ", GLRenderer, ')');
 
         // Under the standard filtering rules for cubemaps, filtering does not work across faces of the cubemap. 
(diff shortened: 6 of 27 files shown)


DiligentGraphics:
Hi!
Linux build fails because some of the include files (xlib I think) `#define None` that creates a conflict.
Mac build apparently fails because cocoa headers can only be used when compiling objective c files.
Windows build fails because clang-format validator errors. Format validation is at this point only automated on Windows. If you generate files with `-D DILIGENT_BUILD_TESTS=TRUE`, there will be targets that automatically validate format, which also run as part of all build.

DiligentGraphics:
Build problems on Linux and Mac is why I don't like to include system headers in public interface.
I suggest all system headers should be removed and forward declarations used if possible, or `void*`.

jayrulez:
I updated the files to use void* and squashed all the fixup commits for a cleaner history.

DiligentGraphics:
Thanks! I will merge this PR little later, have to do something else this week.

jayrulez:
@DiligentGraphics 

Are you okay with changing 
```
    struct EngineGLCreateInfo : public EngineCreateInfo
    {
        /// Native window wrapper
        NativeWindow NativeWindow;
    };
```

to

```
    struct EngineGLCreateInfo : public EngineCreateInfo
    {
        /// Native window wrapper
        NativeWindow NativeWnd;
    };
```

 to fix the linux build?

or maybe rename the NativeWindow type to PlatformWindow?

jayrulez:
I'm not very experienced with C++ so it was after the last commit that I realized I could have just used the SRO here:
```
struct EngineGLCreateInfo : public EngineCreateInfo
    {
        /// Native window wrapper
        ::NativeWindow NativeWindow;
    };
```

But I believe the last commit (the rename) is still okay.

Graphics/GraphicsEngine/interface/GraphicsTypes.h
```diff
@@ -35,6 +35,7 @@
 #include "../../../Primitives/interface/BasicTypes.h"
 #include "../../../Primitives/interface/DebugOutput.h"
 #include "../../../Primitives/interface/FlagEnum.h"
+#include "../../../Platforms/interface/NativeWindowHandle.hpp"
```
DiligentGraphics: `.h` and `.hpp` extensions are used to clearly indicate which files can be used in both c and c++, and which are c++ only. Native window handle will be part of the core API and is compatible with both c and c++, so it should have .h extension

Platforms/Android/interface/AndroidNativeWindowHandle.hpp
```diff
@@ -0,0 +1,35 @@
...
+ *  out of the use or inability to use the software (including but not limited to damages 
+ *  for loss of goodwill, work stoppage, computer failure or malfunction, or any and 
+ *  all other commercial damages or losses), even if such Contributor has been advised 
+ *  of the possibility of such damages.
+ */
+
+#pragma once
+
+#include "../../Primitives/interface/CommonDefinitions.h"
+
```
DiligentGraphics: Please enclose this in `DILIGENT_BEGIN_NAMESPACE`/`DILIGENT_END_NAMESPACE`

Platforms/Android/interface/AndroidNativeWindowHandle.hpp
```diff
@@ -0,0 +1,35 @@
...
+struct AndroidNativeWindowHandle
+{
+    void* pWindowHandle DEFAULT_INITIALIZER(nullptr);
```
DiligentGraphics: Since we are using void* for all handles, let's give them a more descriptive names: `pAWindow`

Platforms/Apple/interface/IOSNativeWindowHandle.hpp
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+#pragma once
+
+#include "../../Primitives/interface/CommonDefinitions.h"
+
+struct IOSNativeWindowHandle
+{
+    void* pWindowHandle DEFAULT_INITIALIZER(nullptr);
+};
```
DiligentGraphics: `pNSView`

Platforms/Apple/interface/MacOSNativeWindowHandle.hpp
```diff
@@ -0,0 +1,35 @@
...
+ *  of the possibility of such damages.
+ */
+
+#pragma once
+
+#include "../../Primitives/interface/CommonDefinitions.h"
+
+struct MacOSNativeWindowHandle
+{
+    void* pWindowHandle DEFAULT_INITIALIZER(nullptr);
```
DiligentGraphics: `pNSView`

Platforms/Linux/interface/LinuxNativeWindowHandle.hpp
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+#pragma once
+
+#include "../../Primitives/interface/CommonDefinitions.h"
+
+struct LinuxNativeWindowHandle
+{
+    void* pDisplay      DEFAULT_INITIALIZER(nullptr);
+    void* pWindowHandle DEFAULT_INITIALIZER(nullptr);
```
DiligentGraphics: `pWindow`

Platforms/UWP/interface/UWPNativeWindowHandle.hpp
```diff
@@ -0,0 +1,35 @@
...
+ *  of the possibility of such damages.
+ */
+
+#pragma once
+
+#include "../../Primitives/interface/CommonDefinitions.h"
+
+struct UWPNativeWindowHandle
+{
+    void* pWindowHandle DEFAULT_INITIALIZER(nullptr);
```
DiligentGraphics: `pWindow`, I guess

Platforms/Win32/interface/Win32NativeWindowHandle.hpp
```diff
@@ -0,0 +1,35 @@
...
+ *  of the possibility of such damages.
+ */
+
+#pragma once
+
+#include "../../Primitives/interface/CommonDefinitions.h"
+
+struct Win32NativeWindowHandle
+{
+    void* pWindowHandle DEFAULT_INITIALIZER(nullptr);
```
DiligentGraphics: `hWnd`

README.md
```diff
@@ -205,9 +205,9 @@ attaches to the one initialized by the app. An example of the engine initializat
 
 On MacOS, Diligent Engine supports OpenGL and Vulkan backends. Initialization of GL context on MacOS is
 performed by the application, and the engine attaches to the context created by the app; see
-[GLView.m](https://github.com/DiligentGraphics/DiligentEngine/blob/master/Common/NativeApp/Apple/Source/Classes/OSX/GLView.m)
+[GLView.m](https://github.com/DiligentGraphics/DiligentTools/tree/master/NativeApp/Apple/Source/Classes/OSX/GLView.m)
```
DiligentGraphics: Thanks, did not know these were broken. Could you submit this a separate PR?

Graphics/GraphicsEngine/interface/GraphicsTypes.h
```diff
@@ -1318,21 +1319,11 @@ struct EngineCreateInfo
...
-    /// * On Android platform, this is a pointer to the native window (ANativeWindow*)
-    /// * On Linux, this is the native window (Window)
-    void* pNativeWndHandle DEFAULT_INITIALIZER(nullptr);
-
-#if PLATFORM_LINUX
-    /// For linux platform only, this is the pointer to the display
-    void* pDisplay         DEFAULT_INITIALIZER(nullptr);
-#endif
+	/// Native window wrapper
+	NativeWindowHandle NativeWindow;
```
DiligentGraphics: Let's just call it `NativeWindow Window`

Platforms/Apple/interface/MacOSNativeWindowHandle.hpp
```diff
@@ -0,0 +1,35 @@
...
+ *  or consequential damages of any character arising as a result of this License or 
+ *  out of the use or inability to use the software (including but not limited to damages 
+ *  for loss of goodwill, work stoppage, computer failure or malfunction, or any and 
+ *  all other commercial damages or losses), even if such Contributor has been advised 
+ *  of the possibility of such damages.
+ */
+
+#pragma once
+
+#include "../../Primitives/interface/CommonDefinitions.h"
```
DiligentGraphics: Should be `../../../` This is why Linux/Mac build fails.

Platforms/Apple/interface/MacOSNativeWindowHandle.hpp
jayrulez: MacOS uses a NSWindow so should we not just use pWindow here? or pNSWindow to be fore specific?

Graphics/GraphicsEngine/interface/GraphicsTypes.h
jayrulez: Should I rename the NativeWindowHandle.h file to NativeWindow.h?

README.md
jayrulez: Yes, I can submit these fix in another PR. I believe there were other broken links too. I will try to find and fix them all.

Platforms/interface/NativeWindowHandle.h
```diff
@@ -32,33 +32,63 @@
 #if PLATFORM_WIN32
 
 #    include "../Win32/interface/Win32NativeWindowHandle.hpp"
+
+DILIGENT_BEGIN_NAMESPACE(Diligent)
+
 typedef struct Win32NativeWindowHandle NativeWindowHandle;
 
+DILIGENT_END_NAMESPACE // namespace Diligent
```
jayrulez: Ideally, I want to get rid of these duplicate namespace declaration in this file. Maybe I should just move the typedef to the platform specific file and just include it here?

DiligentGraphics: Ah, OK. Note that the rest of the platform-specific classes not being in a namespace is a mistake which I am planning to fix at some point.
I would structure the file as follows (and then fix the other headers similar later):

```cpp

#if PLATFORM_WIN32
#    include "../Android/interface/Win32NativeWindow.h"
...
#endif

DILIGENT_BEGIN_NAMESPACE(Diligent)

#if PLATFORM_WIN32
typedef struct Win32NativeWindow NativeWindow;
...
#endif

DILIGENT_END_NAMESPACE
```

There will be duplicated `#if-#elseif` struture, but this is probably a better way then having  multiple `BEGIN_NAMESPACE`/`END_NAMESPACE` sections

Platforms/Apple/interface/MacOSNativeWindowHandle.hpp
DiligentGraphics: +1 for `pNSWindow`

Graphics/GraphicsEngine/interface/GraphicsTypes.h
DiligentGraphics: Yes please

jayrulez:
@DiligentGraphics if everything looks good then I'll squash into 1 commit for a cleaner history but I want your approval first as I believe squashing will cause us to lose the review history.

Tests/DiligentCoreAPITest/include/TestingEnvironment.hpp
```diff
@@ -12,7 +12,7 @@
  *  distributed under the License is distributed on an "AS IS" BASIS,
  *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  *  See the License for the specific language governing permissions and
- *  limitations under the License.
+ *  limitations under the License.hWnd
```
DiligentGraphics: Typo has sneaked

DiligentGraphics:
Looks good to me!

ReleaseHistory.md
```diff
@@ -8,7 +8,7 @@
 
 ### API Changes
 
-* Added `NativeWindowHandle` wrapper and replaced `pNativeWndHandle` and `pDisplay` members with it in `EngineGLCreateInfo` (API Version 2400XX)
+* Added `NativeWindow` wrapper and replaced `pNativeWndHandle` and `pDisplay` members with it in `EngineGLCreateInfo` (API Version 2400XX)
```
DiligentGraphics: Please update this to 240053, as well as [here](https://github.com/DiligentGraphics/DiligentCore/blob/master/Graphics/GraphicsEngine/interface/APIInfo.h#L33). I also updated the file with C interface that I forgot to mentiond.

jayrulez:
Should be ready now. Thanks for the review and patience.

DiligentGraphics:
This PR is merged now: https://github.com/DiligentGraphics/DiligentCore/commit/dadb44ef46cb83a957507a1d2d5d20986b3cec22
I made few changes which confused GitHub, so I the PR did not close automatically.
Thank you for help!

PR #125, opened 2020-04-01
Add CMake option to statically link with MoltenVK

dolphineye:
Hi,

We followed the documentation to add MoltenVK's dylib into our Xcode project and that worked like a charm 👍 

Yet, since our project has a particular setup, and that we build a framework which is then embedded into a final Xcode application, we want to ship MoltenVK with our framework. This, unless mistaken, seems not easily feasible as of today.

We figured that linking with MoltenVK's static library actually does the job pretty well, and might be an acceptable alternative in our situation.

So, in this pull request, I'm adding an optional `DILIGENT_MOLTENVK_LINK_STATIC` option to indicate that we wish to statically link against MoltenVK. Hence this should not impact existing projects. Possibly, this option might be of service in other projects than ours.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,7 +130,13 @@ elseif(PLATFORM_IOS)
         # iOS.toolchain.cmake currently configures cmake to only search iOS SDK for libraries,
         # thus find_library will be unable to find MoltenVK in VulkanSDK
         # find_library(MoltenVK_LIBRARY MoltenVK PATHS "${VULKAN_SDK}/MoltenVK/iOS/dynamic")
-        set(MoltenVK_LIBRARY "${VULKAN_SDK}/MoltenVK/iOS/dynamic/libMoltenVK.dylib" CACHE INTERNAL "MoltenVK library")
+        option(DILIGENT_MOLTENVK_LINK_STATIC "Link with MoltenVK static library" OFF)
+        if(${DILIGENT_MOLTENVK_LINK_STATIC})
+            set(MoltenVK_LIBRARY "${VULKAN_SDK}/MoltenVK/iOS/static/libMoltenVK.a" CACHE INTERNAL "MoltenVK library")
+        else()
+            set(MoltenVK_LIBRARY "${VULKAN_SDK}/MoltenVK/iOS/dynamic/libMoltenVK.dylib" CACHE INTERNAL "MoltenVK library")
+        endif()
+
         if(EXISTS ${MoltenVK_LIBRARY})
             set(VULKAN_SUPPORTED TRUE CACHE INTERNAL "Vulkan is enabled through MoltenVK on iOS platform")
         else()


TheMostDiligent:
Hi! Thanks for submitting the PR.
Did you try to run samples/tutorials in both debug and release modes? Does everything work fine?

This may be little bit more tricky than it appears. In fact, linking statically was the first thing I tried, but there was a very nasty run-time crash in SPIRV-Cross third party library. I even thought that [was a bug](https://github.com/KhronosGroup/SPIRV-Cross/issues/902), however it turned out that there was an issue: both MoltenVK and Diligent link with SPIRV-Cross. When linking everything together, there was some linking issue. No duplicate symbols errors, but, apparently the linker was taking some functions from one implementation and other from the second implementation causing run-time mismatch.
Fortunately, as a response to my issue, SPIRV-Cross team added an option to [override the namespace](https://github.com/KhronosGroup/SPIRV-Cross/issues/902#issuecomment-477641845).

So I guess I can merge your PR and the fix the namespace.

dolphineye:
Hi,

Actually I have not tested the samples & tutorials with this option enabled. Which is a very good point, actually ;)

The symbol collision you mentioned — despite the absence of linker errors — makes sense to me. We experienced similar issues in a past project, and using different namespaces fixed the error. But it can get very nasty and hard to debug.

Overriding the SPIRV-Cross namespace is definitely an idea worth exploring! I suggest we give this a shot. If everything's all good, actually, maybe it's no longer worth keep dynamic linking for MoltenVK.

TheMostDiligent:
I was reading through my issue and apparently MoltenVK team already fixed this problem by namespacing SPIRV-Cross in [their project](https://github.com/KhronosGroup/MoltenVK/blob/master/MoltenVK/MoltenVK.xcodeproj/project.pbxproj#L1211), so everything should be good I think. It still may make sense to namespace SPIRV-Cross to be extra safe though.

dolphineye:
Yeah. And testing the samples & tutorials to make sure everything's OK.

And so btw, worth keeping dynamic linking for MoltenVK you think?

TheMostDiligent:
I would stick to dynamic linking. What issue do you have with it?

dolphineye:
> I would stick to dynamic linking. What issue do you have with it?

None :)

TheMostDiligent:
But you said that shipping MoltenVk with an application is not easily feasible. What is the problem?

dolphineye:
> But you said that shipping MoltenVk with an application is not easily feasible. What is the problem?

To give you some context. We have an app, and this app uses different internal standalone frameworks. One of these frameworks uses Diligent Engine.

If we use dynamic link, we must modify the app's project to reference the MoltenVK library as described in the docs. This works pretty well, but this forces our app to have knowledge of a framework's internals, which is not ideal. Ideally, we'd like our framework to embed MoltenVK's dylib, but I'm not of whether it is possible, or how. Hence we figured using static linking would be our best option so far.

TheMostDiligent:
Yes, I don't know if it is possible to embed a dylib into framework. Probably not. The dylib must be code-signed, so likely it must be added to the main project.

dolphineye:
> Yes, I don't know if it is possible to embed a dylib into framework. Probably not. The dylib must be code-signed, so likely it must be added to the main project.

Best I could find was some script bloat that amends the framework, resign everything. Mhhh.

TheMostDiligent:
I was thinking about this PR. Would it be more flexible, if instead of adding `DILIGENT_MOLTENVK_LINK_STATIC` option, you just set `MoltenVK_LIBRARY` itself when configuring CMake, and then CMake script will check if it is already defined and simply keep i?

dolphineye:
> I was thinking about this PR. Would it be more flexible, if instead of adding `DILIGENT_MOLTENVK_LINK_STATIC` option, you just set `MoltenVK_LIBRARY` itself when configuring CMake, and then CMake script will check if it is already defined and simply keep it?

Good idea. Let me try that and eventually update this PR accordingly.

TheMostDiligent:
I submitted a [few](https://github.com/DiligentGraphics/DiligentCore/commit/bc2b65e2bef87cbdfb8b1c0a3b67246515bfc9e6) [changes](https://github.com/DiligentGraphics/DiligentCore/commit/72eef53a0d290568b61dee06d4739b58325f3b74) to enable this, so now you should be able to define `MoltenVK_LIBRARY` variable to specify the path to any library. I also updated the [instructions in the readme](https://github.com/DiligentGraphics/DiligentEngine#configuring-vulkan-build-environment-1).
Let me know how this works.

dolphineye:
Busy week, I will test it this week end I think. Thanks!

dolphineye:
Hey,

I have done some testing: everything looks good so far.
I guess we can close this PR.

Thanks for helping!

PR #128, opened 2020-04-21
Fallback to GLX_MESA_swap_control if GLX_EXT_swap_control isn't supported

Dinolek:
Tested with Radeon RX 570 Series (POLARIS10, DRM 3.35.0, 5.4.33-3-lts, LLVM 10.0.0)

### DIFF
--- a/Graphics/GraphicsEngineOpenGL/src/GLContextLinux.cpp
+++ b/Graphics/GraphicsEngineOpenGL/src/GLContextLinux.cpp
@@ -166,6 +166,12 @@ void GLContext::SwapBuffers(int SwapInterval)
         {
             glXSwapIntervalEXT(display, wnd, SwapInterval);
         }
+#    if GLX_MESA_swap_control
+        else if (glXSwapIntervalMESA != nullptr)
+        {
+            glXSwapIntervalMESA(SwapInterval);
+        }
+#    endif
 #endif
         glXSwapBuffers(display, wnd);
     }


TheMostDiligent:
Thanks for the PR!
I should say that this construct is very hard to read. It took me good five minutes to understand how it works. I think that the following should work better:

```cpp
#if GLX_EXT_swap_control
        if (glXSwapIntervalEXT != nullptr)
        {
            glXSwapIntervalEXT(display, wnd, SwapInterval);
        }
#elif GLX_MESA_swap_control
        if (glXSwapIntervalMESA != nullptr)
        {
            glXSwapIntervalMESA(SwapInterval);
        }
#endif
```

It does not check both extensions at run time, which I think is probably not necessary. Or is this you actual problem?  In this case I would suggest this way:

```cpp
#if GLX_EXT_swap_control
        if (glXSwapIntervalEXT != nullptr)
        {
            glXSwapIntervalEXT(display, wnd, SwapInterval);
        }
#    if GLX_MESA_swap_control
        else if (glXSwapIntervalMESA != nullptr)
        {
            glXSwapIntervalMESA(SwapInterval);
        }
#    endif
#elif GLX_MESA_swap_control
        if (glXSwapIntervalMESA != nullptr)
        {
            glXSwapIntervalMESA(SwapInterval);
        }
#endif
```
Sometimes some code duplication may not be a bad thing

Dinolek:
Both GLX_EXT_swap_control and GLX_MESA_swap_control are always enabled in glxew.h
The problem is runtime checking.

Graphics/GraphicsEngineOpenGL/src/GLContextLinux.cpp
```diff
@@ -166,12 +166,14 @@ void GLContext::SwapBuffers(int SwapInterval)
...
-#if GLX_MESA_swap_control
-            if (glXSwapIntervalMESA != nullptr)
+#    if GLX_MESA_swap_control
+        else if (glXSwapIntervalMESA != nullptr)
+        {
+            glXSwapIntervalMESA(SwapInterval);
+        }
+#    endif
+#elif GLX_MESA_swap_control
+        if (glXSwapIntervalMESA != nullptr)
```
TheMostDiligent, line 176: Will this not make both `glXSwapIntervalEXT` and `glXSwapIntervalMESA ` being called if they both loaded at run time? I believe this branch is not needed at all since, like you said, both extensions are always statically enabled in GLEW

Dinolek: No?
It's #elif(else if),

Graphics/GraphicsEngineOpenGL/src/GLContextLinux.cpp
```diff
@@ -172,11 +172,6 @@ void GLContext::SwapBuffers(int SwapInterval)
             glXSwapIntervalMESA(SwapInterval);
         }
 #    endif
```
TheMostDiligent, line 174: Looks like `GLX_MESA_swap_control` should not be nested into `GLX_EXT_swap_control`:

```cpp
#if GLX_EXT_swap_control
        if (glXSwapIntervalEXT != nullptr)
        {
            glXSwapIntervalEXT(display, wnd, SwapInterval);
        }
#elif GLX_MESA_swap_control
        if (glXSwapIntervalMESA != nullptr)
        {
            glXSwapIntervalMESA(SwapInterval);
        }
#endif
```

Could you please also merge the changes into one commit to not clobber the history?

Dinolek: With this it's never going to use glXSwapIntervalMESA, because GLX_EXT_swap_control is always enabled.

TheMostDiligent: Ah, Ok. I now see that you version is correct while my is not. Can you merge the history?

PR #130, opened 2020-05-01
Use export map to restrict visible lib symbols under MinGW.

logzero:
This deals with the duplicate symbols issue.
https://github.com/DiligentGraphics/DiligentEngine/issues/31

### DIFF
--- a/Graphics/GraphicsEngineOpenGL/CMakeLists.txt
+++ b/Graphics/GraphicsEngineOpenGL/CMakeLists.txt
@@ -217,6 +217,15 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
         # Disallow missing direct and indirect dependencies to enssure that .so is self-contained
         LINK_FLAGS "-Wl,--no-undefined -Wl,--no-allow-shlib-undefined"
     )
+    if(PLATFORM_WIN32)
+        # Restrict export to GetEngineFactoryOpenGL
+        file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/export.map
+            "{ global: *GetEngineFactoryOpenGL*; local: *; };"
+        )
+        set_property(TARGET Diligent-GraphicsEngineOpenGL-shared APPEND_STRING PROPERTY
+            LINK_FLAGS " -Wl,--version-script=export.map"
+        )
+    endif()
 endif()
 
 target_link_libraries(Diligent-GraphicsEngineOpenGL-static
--- a/Graphics/GraphicsEngineVulkan/CMakeLists.txt
+++ b/Graphics/GraphicsEngineVulkan/CMakeLists.txt
@@ -241,6 +241,15 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
         # Disallow missing direct and indirect dependencies to enssure that .so is self-contained
         LINK_FLAGS "-Wl,--no-undefined -Wl,--no-allow-shlib-undefined"
     )
+    if(PLATFORM_WIN32)
+        # Restrict export to GetEngineFactoryVk
+        file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/export.map
+            "{ global: *GetEngineFactoryVk*; local: *; };"
+        )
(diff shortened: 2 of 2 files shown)


Graphics/GraphicsEngineOpenGL/CMakeLists.txt
```diff
@@ -213,10 +213,20 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
...
-        LINK_FLAGS "-Wl,--no-undefined -Wl,--no-allow-shlib-undefined"
-    )
+    if(PLATFORM_WIN32)
+        file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/export.map
+            "{ global: *GetEngineFactoryOpenGL*; local: *; };"
+        )
+        set_target_properties(Diligent-GraphicsEngineOpenGL-shared PROPERTIES
+            # Disallow missing direct and indirect dependencies to enssure that .so is self-contained
+            LINK_FLAGS "-Wl,--no-undefined -Wl,--no-allow-shlib-undefined -Wl,--version-script=export.map"
+        )
```
TheMostDiligent, line 223: Will this work if you just have this win32-specific branch to eliminate duplication:
```
if(PLATFORM_WIN32)
    # Please write a comment about what this does
    file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/export.map
          "{ global: *GetEngineFactoryOpenGL*; local: *; };"
    )
    set_target_properties(Diligent-GraphicsEngineOpenGL-shared PROPERTIES
           LINK_FLAGS "-Wl,--version-script=export.map"
    )
endif()
```

Graphics/GraphicsEngineVulkan/CMakeLists.txt
```diff
@@ -237,10 +237,20 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
...
     )
 elseif (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
-    set_target_properties(Diligent-GraphicsEngineVk-shared PROPERTIES
-        # Disallow missing direct and indirect dependencies to enssure that .so is self-contained
-        LINK_FLAGS "-Wl,--no-undefined -Wl,--no-allow-shlib-undefined"
-    )
+    if(PLATFORM_WIN32)
+        file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/export.map
+            "{ global: *GetEngineFactoryVk*; local: *; };"
+        )
```
TheMostDiligent, line 243: Same as above

logzero:
My CMake skills are pretty basic. Appending to LINK_FLAGS seems to work as expected.

TheMostDiligent:
I still don't quite understand what your changes do and why they were necessary, but they certainly fixed linking errors - thanks!

TheMostDiligent:
As to xinput errors - they come from imgui, will think what to do about them

TheMostDiligent:
I updated imgui to the latest version and it fixed the problem with XInput, but introduced another problem with `SetProcessDPIAware`:
```
imgui_impl_win32.cpp:408:5: error: 'SetProcessDPIAware' was not declared in this scope
```
I don't know why MinGW is complaining, this function seems to be [around since 2008](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setprocessdpiaware).

TheMostDiligent:
So commenting out [this line](https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_win32.cpp#L408) makes everything work. I opened an [issue vs DearImgui](https://github.com/ocornut/imgui/issues/3183).

logzero:
> 
> 
> I still don't quite understand what your changes do and why they were necessary, but they certainly fixed linking errors - thanks!

My understanding is that the visibility attributes are a compiler thing, which the compiler has to communicate to the linker. I guess mingw does not support them correctly?

-Wl,--version-script on the other side is telling the linker explicitely what to do.

logzero:
Btw I used APPEND because set_target_properties override previous settings. Wasn't sure if that was intended.

TheMostDiligent:
> set_target_properties override previous settings

Thanks for pointing out at this, I honestly was 100% sure that it appends link flags.

PR #133, opened 2020-05-05
VK_LAYER_LUNARG_standard_validation deprecated

markusgod:
https://vulkan.lunarg.com/doc/view/1.0.57.0/windows/validation_layers.html VK_LAYER_LUNARG_standard_validation is deprecated use VK_LAYER_KHRONOS_validation instead.


TheMostDiligent:
Thanks for the PR! As a matter of fact, in 1.1.135.0 SDK the layers seem to load, but do nothing with the old string.

PR #137, opened 2020-05-08
Small fixes.

markusgod:


Graphics/GraphicsEngineOpenGL/src/FBOCache.cpp
```diff
@@ -67,7 +67,6 @@ std::size_t FBOCache::FBOCacheKeyHashFunc::operator()(const FBOCacheKey& Key) co
     if (Key.Hash == 0)
     {
         std::hash<TextureViewDesc> TexViewDescHasher;
-        Key.Hash = 0;
         HashCombine(Key.Hash, Key.NumRenderTargets);
```
TheMostDiligent, line 70: I would prefer to keep this way, it doesn't hurt. HashCombine modifies its first argument, so it makes it clear what value it has originally.

markusgod: I'll revert change than.

PR #142, opened 2020-05-12
Fixing typos in matrix4x4, matrix3x3, matrix2x2 operator==

BugzTroll:
Hello! I noticed a typo in the matrices comparison operators in BasicMath.hpp. Here's the fix!


TheMostDiligent:
Good catch, thanks!

PR #150, opened 2020-08-02
Fix ARM64 compilation.

tombish:
MSVC _CountOneBits intrinsics undefined for ARM64.


tombish:
Method taken from BSD licenced ANGLE project.
https://chromium.googlesource.com/angle/angle/+/refs/heads/master/src/common/mathutil.h

PR #154, opened 2020-08-07
Validate format won't crash if it cannot load the provided binary

dpellegr:
Fix for https://github.com/DiligentGraphics/DiligentEngine/issues/79

### DIFF
--- a/BuildTools/FormatValidation/validate_format_linux.sh
+++ b/BuildTools/FormatValidation/validate_format_linux.sh
@@ -1,7 +1,35 @@
 #!/bin/bash
-python clang-format-validate.py --clang-format-executable ./clang-format_linux_10.0.0 \
--r ../../Common ../../Graphics ../../Platforms ../../Primitives ../../Tests \
---exclude ../../Graphics/HLSL2GLSLConverterLib/include/GLSLDefinitions.h \
---exclude ../../Graphics/HLSL2GLSLConverterLib/include/GLSLDefinitions_inc.h \
---exclude ../../Graphics/GraphicsEngineVulkan/shaders/GenerateMipsCS_inc.h \
---exclude ../../Tests/DiligentCoreAPITest/assets/*
+
+## Set the shipped BIN
+BIN="./clang-format_linux_10.0.0"
+## ...or use the find command to avoid maintaining
+#BIN=$(find . -name 'clang-format_linux*')
+
+## Set to anything but 0 to force skip format validation
+SKIP=0
+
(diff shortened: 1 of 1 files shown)


TheMostDiligent:
Thanks for the patch!
I have two questions/suggestions:
- Probably we can also check if the version of the system binary is the same as expected, and don't print warning in this case.
- How `SKIP` variable can be provided to the script? If it is expected to be a system-level variable, I would suggest renaming it to `DILIGENT_SKIP_FORMAT_VALIDATION`, which is little verbose, but unambiguous

dpellegr:
* Yes, but the only way for me to get it is by parsing the file name, is this ok?
* If you can call the script with `$SKIP` as an (optional) argument, then I can easily pull it in from the outside. I thought about it as a last resort for the user to proceed, but it can be made a system variable.

TheMostDiligent:
>  Yes, but the only way for me to get it is by parsing the file name, is this ok?

Just having it defined as a variable in the script would be OK. The binary name can then be constructed using this version.

>  How `SKIP` variable can be provided to the script? If it is expected to be a system-level variable, I would suggest renaming it to DILIGENT_SKIP_FORMAT_VALIDATION, which is little verbose, but unambiguous

The script is executed from the build rule created by CMake as part of the build and there is no direct way to pass an argument to it, which is why I was thinking about global variable. Alternatively, I can add the same CMake option, and then` SKIP` in the script will not be needed

dpellegr:
Ok, I'll add the version in the script and use it to build the name.

Concerning the second point I think that your latter proposal is the best. I will drop the variable in the script.

TheMostDiligent:
That looks great, thanks!
Do you think it makes sens to follow [this suggestion from Codacy](https://app.codacy.com/manual/DiligentGraphics/DiligentCore/pullRequest?bid=19607018&prid=5979505&utm_campaign=new_commit&utm_medium=Email&utm_source=Internal&)?

dpellegr:
I ended up parsing the file name, as I needed the same logic to clean the output of `clang-format --version`.
The advantage is that as long as you keep the binary name as: `clang-format_linux_X.Y.Z` there is no further need to edit the script on updates.

The suggestion was ok, it is always safer to quote.

dpellegr:
I am happy to see this merged!
Could you please also update the submodule in https://github.com/DiligentGraphics/DiligentEngine
Thank you!

TheMostDiligent:
> Could you please also update the submodule in https://github.com/DiligentGraphics/DiligentEngine

Let's fix other submodules and then I will update all of them.

PR #162, opened 2020-09-16
Mesh shader fix

azhirnov:


Graphics/GraphicsEngineD3D11/include/PipelineStateD3D11Impl.hpp
```diff
@@ -142,9 +142,9 @@ class PipelineStateD3D11Impl final : public PipelineStateBase<IPipelineStateD3D1
     // SRB memory allocator must be defined before the default shader res binding
     SRBMemoryAllocator m_SRBMemAllocator;
 
-    Int8 m_ResourceLayoutIndex[6] = {-1, -1, -1, -1, -1, -1};
+    std::array<Int8, NUM_SHADER_TYPES> m_ResourceLayoutIndex;
```
TheMostDiligent, line 145: We can have at most 5 different shaders at in one pipeline. Now there are 3 orthogonal pipeline types. With ray tracing, there will be another pipeline type and 5 more shader stages.
A proper solution is to add pipeline type parameter to GetShaderTypeIndex function to make indices overlap.
I will work on this.

Graphics/GraphicsEngineOpenGL/include/GLPipelineResourceLayout.hpp
```diff
@@ -328,9 +331,9 @@ class GLPipelineResourceLayout
 /*34*/ OffsetType m_ImageOffset         = 0;
 /*36*/ OffsetType m_StorageBufferOffset = 0;
 /*38*/ OffsetType m_VariableEndOffset   = 0;
-/*40*/ std::array<Int8, 6> m_ProgramIndex = {{-1, -1, -1, -1, -1, -1}};
-/*46*/ Uint8      m_NumPrograms         = 0;
-/*48*/
+/*40*/ std::array<Int8, NUM_SHADER_TYPES> m_ProgramIndex;
```
TheMostDiligent, line 334: FYI, the size of the struct will be 56 bytes, which is why I have the offset as a hint

Graphics/GraphicsEngineD3D11/include/PipelineStateD3D11Impl.hpp
azhirnov: Pipeline type is not needed in GetShaderTypeIndex:
```cpp
case SHADER_TYPE_PIXEL: return 0;
case SHADER_TYPE_VERTEX:
case SHADER_TYPE_AMPLIFICATION: return 1;
case SHADER_TYPE_GEOMETRY:
case SHADER_TYPE_MESH: return 2;
...
```

TheMostDiligent: Yes, I was thinking about that too. I may add the pipeline type for debug purposes, just to make sure that e.g. mesh shader is not requested for compute pipeline.

Graphics/GraphicsEngineD3D12/include/PipelineStateD3D12Impl.hpp
```diff
@@ -129,7 +129,7 @@ class PipelineStateD3D12Impl final : public PipelineStateBase<IPipelineStateD3D1
     ShaderResourceCacheD3D12*   m_pStaticResourceCaches  = nullptr;
     ShaderVariableManagerD3D12* m_pStaticVarManagers     = nullptr;
     // Resource layout index in m_ResourceLayouts[] array for every shader stage
-    Int8 m_ResourceLayoutIndex[6] = {-1, -1, -1, -1, -1, -1};
+    std::array<Int8, NUM_SHADER_TYPES> m_ResourceLayoutIndex;
```
TheMostDiligent, line 132: You removed default initialization here, but did not add it to the constructor. This could've resulted in a very hard-to-find issues.

PR #169, opened 2020-10-15
GraphicsPipelineCreateInfo and ComputePipelineCreateInfo instead of PipelineCreateInfo

azhirnov:


Common/interface/LinearAllocator.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+    void AddRequiredSize(size_t size, size_t align)
+    {
+        VERIFY(m_pBuffer == nullptr, "Memory already allocated");
+        if (size > 0)
+        {
+            m_RequiredSize = Align(m_RequiredSize, align) + size;
+
+            // Reserve additional space for pointer alignment
+            m_RequiredSize += (align > sizeof(void*) ? align : 0);
```
TheMostDiligent, line 104: I think this should be `+= Align(m_RequiredSize, sizeof(void*)) - m_RequiredSize` to make it aligned to the closest `sizeof(void)` rather than always reserve extra `sizeof(void*)`

Common/interface/LinearAllocator.hpp
```diff
@@ -0,0 +1,208 @@
...
+        m_RequiredSize = 0;
+        return Ptr;
+    }
+
+    void AddRequiredSize(size_t size, size_t align)
+    {
+        VERIFY(m_pBuffer == nullptr, "Memory already allocated");
+        if (size > 0)
+        {
+            m_RequiredSize = Align(m_RequiredSize, align) + size;
```
TheMostDiligent, line 101: I think that `size` should be alined too.

azhirnov: Usually size is already aligned, but I can add align to be sure.
Do I need to align size in allocation?

Common/interface/LinearAllocator.hpp
azhirnov: ok

PR #175, opened 2020-11-16
Ray tracing 6 (final)

azhirnov:


TheMostDiligent:
This is now merged into master. Not sure why GitHub did not automatically recognize it.

PR #191, opened 2021-02-26
Replaced VkFencePool to Timeline Semaphore in VkFenceImpl

MikhailGorobets:
I think the timeline semaphore makes the code cleaner and more DX12-like.


TheMostDiligent:
Hi! Thanks for submitting this PR! Timeline semaphores indeed make things clearer and consistent between Vk and D3D12. However, they are an extension and are only a core feature in Vulkan 1.2. Since we have to support 1.0 devices, we can't remove the old method.

MikhailGorobets:
It is possible to make #ifdef ... #endif for this. If developer wants to use minimum Vulkan version as 1.2

TheMostDiligent:
> Hi. It is possible to make #ifdef ... #endif for this. If developer wants to use minimum Vulkan version as 1.2

This must be a run-time check because the device runtime may not support this feature. There is a new [feature](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#features-timelineSemaphore) that must be checked to see if timeline semaphores are supported by the device.

MikhailGorobets:
> This must be a run-time check because the device runtime may not support this feature. There is a new [feature](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#features-timelineSemaphore) that must be checked to see if timeline semaphores are supported by the device.
I'll add this. I also think we need to think about how to maintain compatibility with VkFencePool when the device doesn't support VK_KHR_timeline_semaphore

TheMostDiligent:
> I also think we need to think about how to maintain compatibility with VkFencePool when the device doesn't support VK_KHR_timeline_semaphore

Like I said, `VkFencePool` method can't be removed and we are likely stuck with it forever. Having two fence implementations adds complexity and leads to possible errors. Unless there is a clear advantage to using timeline semaphores over `VkFence`, which I do not see now, there is no benefit in having two versions.

TheMostDiligent:
Merged in https://github.com/DiligentGraphics/DiligentCore/commit/138fb1fc7b3d29a36048dcef0479ba92a3048234

PR #195, opened 2021-03-29
Documentation addition

jcpulido97:
I noticed that the current online documentation is not updated and doxygen is not quite the greatest experience so I added [m.css](https://github.com/mosra/m.css) to create a more usable version of the doc.


jcpulido97:
I hosted the documentation on my local repo but you should be able to replace the link to yours.

You can check it out [here](https://jcpulido97.github.io/DiligentEngineDoc/html/index.html) 

Preview:
image

Class/Struct navigation:
image

PR #245, opened 2021-05-23
fix: warning C5054 in VS2019

wyom1ng:
Arithmetic between different enum types is deprecated, so a cast is
necessary.

Just because it's not immediately obvious from the code change, the issue occurs in `Graphics/GraphicsEngineVulkan/src/PipelineResourceSignatureVkImpl.cpp` when multiplying `SetId` (`PipelineResourceSignatureVkImpl::DESCRIPTOR_SET_ID`) with the `CACHE_GROUP` enum.

### DIFF
--- a/Common/CMakeLists.txt
+++ b/Common/CMakeLists.txt
@@ -34,6 +34,7 @@ set(INTERFACE
     interface/UniqueIdentifier.hpp
     interface/ValidatedCast.hpp
     interface/CompilerDefinitions.h
+    interface/EnumCast.hpp
 )
 
 set(SOURCE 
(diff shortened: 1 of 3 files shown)


TheMostDiligent:
Which Visual Studio version do you use? We build the engine with `/WX`, so we can't miss any warning.
Besides, a much simpler way to fix the warning would be:
```cpp
const size_t SetId             = VarTypeToDescriptorSetId(Res.VarType);
```

TheMostDiligent:
Fixed in https://github.com/DiligentGraphics/DiligentCore/commit/b357896f6fffd10d5a7572638a5b265e1edd8480

PR #253, opened 2021-07-05
add possibility to disable dxc

AndreyMlashkin:
Some users may not need DirectXShaderCompiler support.
Introduce a cmake flag to disable it

### DIFF
--- a/Graphics/ShaderTools/CMakeLists.txt
+++ b/Graphics/ShaderTools/CMakeLists.txt
@@ -43,7 +43,11 @@ if((PLATFORM_WIN32 AND NOT MINGW_BUILD) OR PLATFORM_UNIVERSAL_WINDOWS OR PLATFOR
     set(DXC_SUPPORTED TRUE)
 endif()
 
-if (DXC_SUPPORTED)
+if(NOT DILIGENT_NO_DIRECT3D11 OR NOT DILIGENT_NO_DIRECT3D12)
+  set(DXC_NEEDED TRUE)
+endif()
+
+if (DXC_SUPPORTED AND DXC_NEEDED)
     list(APPEND INCLUDE include/DXCompiler.hpp)
     list(APPEND SOURCE src/DXCompiler.cpp)
 
@@ -105,7 +109,7 @@ PRIVATE
     ../GraphicsEngine/include
 )
 
-if (DXC_SUPPORTED)
+if (DXC_SUPPORTED AND DXC_NEEDED)
     target_include_directories(Diligent-ShaderTools PUBLIC ../../ThirdParty/DirectXShaderCompiler)
 
     if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
--- a/Graphics/GraphicsEngineVulkan/src/ShaderVkImpl.cpp
+++ b/Graphics/GraphicsEngineVulkan/src/ShaderVkImpl.cpp
@@ -35,7 +35,6 @@
 #include "RenderDeviceVkImpl.hpp"
 #include "DataBlobImpl.hpp"
 #include "GLSLUtils.hpp"
-#include "DXCompiler.hpp"
 #include "ShaderToolsCommon.hpp"
 
 #if !DILIGENT_NO_GLSLANG


TheMostDiligent:
Please rename this option to `DILIGENT_NO_DXC` and move it to the ShaderTools' CMake file.

TheMostDiligent:
Why do you need to disable it, by the way?

AndreyMlashkin:
I need to run diligent in a docker container and I need to keep it as small as possible.

AndreyMlashkin:
Is there a CI problem?
By default, nothing should changed, only if user overrides the flag

azhirnov:
@AndreyMlashkin DirectX shaders uses DXC to get reflection from DXIL, you should surround this code by macros

TheMostDiligent:
> Is there a CI problem?

Yes, DX12 backend expects that DXC is always present and there is a build error. It is expected to be the main compiler for DX12. If you disable it, you will not be able to consume DXIL. None of the next-gen DX12 features (mesh shaders, ray tracing, wave operations, etc.) will not work.

> I need to run diligent in a docker container and I need to keep it as small as possible.

You are disabling one file worth of ~1800 lines of code. Not sure this will give you a ton of space savings.
Note that you are not disabling DXC itself, but only the code that calls. The compiler itself is in the System directory or the SDK.

AndreyMlashkin:
I use vulkan installation from conan, it contains only minimal libs (only vulkanlib actually), there is no dxc compiler at all.
That's why I also want to exclude code, that uses it

Graphics/ShaderTools/CMakeLists.txt
```diff
@@ -39,7 +41,7 @@ if(D3D11_SUPPORTED OR D3D12_SUPPORTED)
     list(APPEND INCLUDE include/ResourceBindingMap.hpp)
 endif()
 
-if((PLATFORM_WIN32 AND NOT MINGW_BUILD) OR PLATFORM_UNIVERSAL_WINDOWS OR PLATFORM_LINUX)
+if(DILIGENT_NO_DXC AND ((PLATFORM_WIN32 AND NOT MINGW_BUILD) OR PLATFORM_UNIVERSAL_WINDOWS OR PLATFORM_LINUX))
```
TheMostDiligent, line 44: Should be `NOT DILIGENT_NO_DXC`

Graphics/ShaderTools/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ cmake_minimum_required (VERSION 3.6)
 
 project(Diligent-ShaderTools CXX)
 
+set(DILIGENT_NO_DXC FALSE CACHE INTERNAL "Use DirectXShaderCompiler compiler")
```
TheMostDiligent, line 5: I don't think this needs to be a cache variable. If it is internal and not an CMake option, there is no need to define this variable here at all.

TheMostDiligent: Besides, this variable defines whether to **disable** using the compiler, not when to use it

TheMostDiligent: Please remove this variable, it is not needed

TheMostDiligent (changes requested):

AndreyMlashkin:
I have tried to compile my changes only in Windows for different flags.

AndreyMlashkin:
@TheMostDiligent , I would kindly ask you to start the CI pipeline

Graphics/ShaderTools/CMakeLists.txt
```diff
@@ -43,7 +43,11 @@ if((PLATFORM_WIN32 AND NOT MINGW_BUILD) OR PLATFORM_UNIVERSAL_WINDOWS OR PLATFOR
     set(DXC_SUPPORTED TRUE)
 endif()
 
-if (DXC_SUPPORTED)
+if(NOT DILIGENT_NO_DIRECT3D11 OR NOT DILIGENT_NO_DIRECT3D12)
+  set(DXC_NEEDED TRUE)
+endif()
```
TheMostDiligent, line 48: This logic is incorrect. On Linux, for example, there is no D3D11 and D3D12 even when neither DILIGENT_NO_DIRECT3D11 nor DILIGENT_NO_DIRECT3D12 is defined, but DXC is still needed.
The correct check is if (D3D11_SUPPORTED or D3D12_SUPPORTED)

Graphics/ShaderTools/CMakeLists.txt
```diff
@@ -43,7 +43,11 @@ if((PLATFORM_WIN32 AND NOT MINGW_BUILD) OR PLATFORM_UNIVERSAL_WINDOWS OR PLATFOR
...
+
+if (DXC_SUPPORTED AND DXC_NEEDED)
```
TheMostDiligent, line 50: This is also incorrect. DXC is used by Vulkan backend even when there is no D3D11 and D3D12. All high-end tutorials (mesh shaders, ray tracing, hybrid rendering) require DXC compiler in Vulkan mode

TheMostDiligent:
I am leaning to close this PR as it does not really provide any benefit. Removing one file from the build is not going to give any significant space savings, but complicates the build configuration by a lot. If you don't have DXC on your system, everything will still run fine unless you try to use `SHADER_COMPILER_DXC`. You need to install ATL support for Visual Studio and the file will compile without errors.

PR #255, opened 2021-07-07
Use 3rd party libs from conan

AndreyMlashkin:
Use opengl and glew libraries from conan.
You can get all 3rd party libs this way, I would see it as a first step.


TheMostDiligent:
We use specific versions of every third-party library to make sure that the project is always in a good state, that it will build [and run fine](https://github.com/DiligentGraphics/DiligentEngine/issues/80#issuecomment-625085606). This also guarantees that very client will get the exact same code and there are not issues caused by mismatch with thrid-party libraries. We will not be switching to package management.

Not to mention that this change breaks the entire build system.

AndreyMlashkin:
I think you misunderstand a concept of a package manager.
Conan uses a specific released version (like 'REQUIRES glew/2.2.0')
or a specific commit for projects without tags like spirv-headers.

AndreyMlashkin:
The change I introduced here does not break anything. It only requires user to have conan installed, all the rest stays the same as before.

TheMostDiligent:
Right now there is a single command that downloads all the source code base required to build the engine:

```
git clone --recursive https://github.com/DiligentGraphics/DiligentEngine.git
```

It does not require any third-party dependencies, all the source code is hosted on GitHub in repositories owned by DG. No need to install anything, no dependencies on other providers. Like I mentioned in another issue, some repositories contain minor changes which makes it not possible to download them as packages. This means that while theoretically it is possible to download some of the third-party libs as packages, others will remain as submodules, which will make third-party handling inconsistent and hard to maintain. 

> The change I introduced here does not break anything

It most definitely [breaks the Windows build on Appveyor](https://ci.appveyor.com/project/DiligentGraphics/diligentcore/builds/39900956). I have not even started the main CI process that builds the engine in 22 different configurations, since the majority of them will certainly be broken too. So most definitely this change does not work out-of-box.

> It only requires user to have conan installed

You don't want to install Visual Studio C++ ATL support, which is mentioned [as a required dependency](https://github.com/DiligentGraphics/DiligentEngine#win32) and instead are adding a new CMake option that disables the component that does not build. Yet you expect everyone else to have conan package manager installed, without which it will not be possible to get the code at all.

I am very open to make the project flexible so that it meets the requirements of different users. That however does not mean that I will be accepting PRs that make significant changes to tailor the needs of one specific project.

If you want to use third-party libraries sourced differently, I am open to discuss how to make this work. One solution I can propose is to check if the target already exists, e.g.:

```cmake
if(NOT TARGET glew-static)
    add_subdirectory(glew)
endif()
```

This way you will be able to download all third-party modules as packages and use them instead of the submoudles. You will only need to `add_subdirectory()` before adding DiligentCore

AndreyMlashkin:
Right now it is quite hard to integrate Diligent to a bigger project, that already has many dependencies and should be built in many different configurations.

Just an example: if you try to override windows runtime (-DCMAKE_CXX_FLAGS_RELEASE="/MT"), compilation will fail, because glslang manages this flag different from all other dependencies in the project. MSVC can't handle runtime mismatch across the libs.

TheMostDiligent:
> because glslang manages this flag different from all other dependencies in the project

I am not sure how downloading glslang as a package will solve this problem - if you download the right commit, it will be exactly the same source.

Diligent allows client build system to [override its settings](https://github.com/DiligentGraphics/DiligentEngine#customizing-build). This however does not apply to the third-party projects currently, but can be changed to do so too.

PR #257, opened 2021-07-20
use ENABLE_RTTI flag the same way as glslang

AndreyMlashkin:
glslang has ENABLE_RTTI flag. I have done the same in Diligent to harmonize work around this flag

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ else()
     set(DILIGENT_CORE_PRO_EXISTS FALSE)
 endif()
 
+option(ENABLE_RTTI "Enables RTTI" OFF)
 
 set(PLATFORM_WIN32 FALSE CACHE INTERNAL "")
 set(PLATFORM_UNIVERSAL_WINDOWS FALSE CACHE INTERNAL "")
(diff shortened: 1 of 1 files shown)


TheMostDiligent:
Diligent requires RTTI in debug build as it performs run-time type checks. Without RTTI nothing will run in Debug. In release build RTTI is always disabled.

AndreyMlashkin:
Current implementation does not leave a possibility of a lib user to choose, if he needs RTTI in release mode, as, for example glslang.
I have just copied parts of code from glslang to CMake root.

TheMostDiligent:
> Current implementation does not leave a possibility of a lib user to choose, if he needs RTTI in release mode, as, for example glslang.

Like I said, the library uses RTTI in debug mode to perform run-time type checks. Without RTTI all these checks will fail. In release mode these checks are not performed and RTTI is always disabled - there is no reason to enable it.
The user don't have an option to disable features that are required for the library to work properly.

> I have just copied parts of code from glslang to CMake root.

What works for one project does not necessarily work verbatim or is needed at all for another.

AndreyMlashkin:
Could you then explain me, why do you need to disable rtti only for MSVC?

TheMostDiligent:
RTTI should be disabled for other compilers in release configurations too. It is an oversight it is not disabled.

AndreyMlashkin:
of course it breaks debug builds as default value set to OFF in PR

TheMostDiligent:
So what is the value of the PR that expectedly breaks the build? Do you realize that this change will break the build of *all* customers who use the engine?

PR #260, opened 2021-07-31
Fix ubuntu build errors

logankaser:
This fixes a variety of build errors on Ubuntu (clang and gcc), although some would occur on other platforms.

The `std::atomic_int32_t` and `std::atomic_unt64_t` type aliases are optional and were not available for me, so I replaced with `std::atomic<std::int32_t>` and `std::atomic<std::uint64_t>` which are always portable.

I also fixed several unused variable warnings because of code that was only used in debug mode, and
Added a magic `// fall through` comment that is the most portable way to tell compilers to be okay with
a switch case fallthough.

I also added a default virtual destructor to `EngineFactoryBase` (`virtual ~EngineFactoryBase() = default;`) which, it should have anyway, to prevent a warning with my system and compiler mix.

I've tried to keep the impact of these changes as low as possible. I should also mention that I have only been able to compile the `OpenGL` backend so far, (I've had to make my own bazel `BUILD` file, so it has been some effort).
Let me know if you think these changes don't belong upstream, I can always just maintain an internal fork if needed.

Also, at HLSL2GLSLConverterImpl.cpp:1774 I fixed a `-WParentheses` warning, but I'm not sure I interpreted your intent correctly.

### DIFF
--- a/Graphics/HLSL2GLSLConverterLib/src/HLSL2GLSLConverterImpl.cpp
+++ b/Graphics/HLSL2GLSLConverterLib/src/HLSL2GLSLConverterImpl.cpp
@@ -1771,7 +1771,7 @@ void HLSL2GLSLConverterImpl::ConversionStream::ParseSamplers(TokenListType::iter
         else
             ++Token;
     }
-    VERIFY_PARSER_STATE(Token, ScopeDepth == 1 && Token == m_Tokens.end() || ScopeDepth == 0, "Error parsing scope");
+    VERIFY_PARSER_STATE(Token, (ScopeDepth == 1 && Token == m_Tokens.end()) || ScopeDepth == 0, "Error parsing scope");
 }
 
 void ParseImageFormat(const String& Comment, String& ImageFormat)
@@ -3065,7 +3065,7 @@ void HLSL2GLSLConverterImpl::ConversionStream::ProcessFunctionParameters(TokenLi
                     // ^
                     ParamInfo.GSAttribs.PrimType = ShaderParameterInfo::GSAttributes::PrimitiveType::TriangleAdj;
                     ++Token;
-
+                // fall through
                 case TokenType::kw_TriangleStream:
                 case TokenType::kw_PointStream:
                 case TokenType::kw_LineStream:
@@ -3975,6 +3975,7 @@ void HLSL2GLSLConverterImpl::ConversionStream::ProcessHullShaderArguments(TokenL
         quad,
         isoline
     } domain = Domain::undefined;
+    (void)domain;
     enum class Partitioning
     {
         undefined,
@@ -3983,6 +3984,7 @@ void HLSL2GLSLConverterImpl::ConversionStream::ProcessHullShaderArguments(TokenL
         fractional_odd,
         pow2
     } partitioning = Partitioning::undefined;
+    (void)partitioning;
     enum class OutputTopology
     {
         undefined,
@@ -3991,6 +3993,7 @@ void HLSL2GLSLConverterImpl::ConversionStream::ProcessHullShaderArguments(TokenL
         triangle_cw,
         triangle_ccw
     } topology = OutputTopology::undefined;
+    (void)topology;
 
     std::unordered_map<HashMapStringKey, String, HashMapStringKey::Hasher> Attributes;
     ProcessShaderAttributes(Token, Attributes);
--- a/Common/interface/UniqueIdentifier.hpp
+++ b/Common/interface/UniqueIdentifier.hpp
@@ -66,7 +66,7 @@ public:
     {
         if (m_ID == 0)
         {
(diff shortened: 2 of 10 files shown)


Graphics/HLSL2GLSLConverterLib/src/HLSL2GLSLConverterImpl.cpp
```diff
@@ -1771,7 +1771,7 @@ void HLSL2GLSLConverterImpl::ConversionStream::ParseSamplers(TokenListType::iter
         else
             ++Token;
     }
-    VERIFY_PARSER_STATE(Token, ScopeDepth == 1 && Token == m_Tokens.end() || ScopeDepth == 0, "Error parsing scope");
+    VERIFY_PARSER_STATE(Token, (ScopeDepth == 1 && Token == m_Tokens.end()) || ScopeDepth == 0, "Error parsing scope");
```
logankaser, line 1774: I am not sure if I interpreted this right

TheMostDiligent:
What is your build set up that produces all the these issues? We build the engine in 23 different configurations with werror and all run clean. On Ubuntu in particular we build with both clang and gcc.
There are some variables that are used in debug checks and are unused in release builds. There is a lot of them, so the unused variable warning is disabled in release, but not in debug build.

logankaser:
> What is your build set up that produces all the these issues? We build the engine in 23 different configurations with werror and all run clean. On Ubuntu in particular we build with both clang and gcc.
> There are some variables that are used in debug checks and are unused in release builds. There is a lot of them, so the unused variable warning is disabled in release, but not in debug build.

I understand if you would rather reject this change, since my situation is a little odd.
I've been porting this project to the bazel build system for use at my job (Zoox), while I've tried to keep things as close to your cmake setup as possible, I don't yet have anything to use different flags for different types(release/debug) of builds (for various reasons that's quite annoying to do in my situation). Some other issues are just compiler and C++ std version related though, for example `std::atomic_int32_t` is not available for me.
If you take nothing else from this patch the `std::atomic_int32_t` change would be the best since I can't fix that by disabling more compiler diagnostics, and I think that is the most likely to affect other people.

TheMostDiligent:
> I've been porting this project to the bazel build system for use at my job (Zoox)

This may work for OpenGL backend, which is the simplest one and does not have many source files or third-party dependencies. When you get to Vulkan though, things may get really complicated.
Can you build the shared libs using CMake and then use them in Bazel?

> I don't yet have anything to use different flags for different types(release/debug) of builds (for various reasons that's quite annoying to do in my situation)

This may be quite an inconvenience. In debug/development Diligent performs a lot of validation and catches virtually all API usage errors printing error messages. In release build all these checks are disabled for performance reasons and invalid API usage will most likely result in a black screen or app crash.

> If you take nothing else from this patch the std::atomic_int32_t change would be the best since I can't fix that by disabling more compiler diagnostics, and I think that is the most likely to affect other people.

Yes, I think fixing atomics makes sense. I can do this since this should be done consistently for all backends, not only OpenGL. For unused variable warning, you should disable it in the compiler setting. Wrapping all debug checks in #if DILIGENT_DEBUG only clubbers the code without any benefit other than silencing the warning.

Graphics/HLSL2GLSLConverterLib/src/HLSL2GLSLConverterImpl.cpp
```diff
@@ -3065,7 +3065,7 @@ void HLSL2GLSLConverterImpl::ConversionStream::ProcessFunctionParameters(TokenLi
                     // ^
                     ParamInfo.GSAttribs.PrimType = ShaderParameterInfo::GSAttributes::PrimitiveType::TriangleAdj;
                     ++Token;
-
+                // fall through
```
TheMostDiligent, line 3068: This is actually a bug...

TheMostDiligent:
I fixed atomics and a couple of other issues. Unused variables should be fixed using the compiler flags.

Graphics/HLSL2GLSLConverterLib/src/HLSL2GLSLConverterImpl.cpp
logankaser: Yeah, that's exactly why my company has that flag on by default, you can make the compiler ignore it if you put a comment like this there.

logankaser:
Seems like you have fixed the parts that make sense.

Graphics/HLSL2GLSLConverterLib/src/HLSL2GLSLConverterImpl.cpp
TheMostDiligent: I honestly don't understand why it didn't get caught. We [use `-Werror -Wall -Wextra` plus few other additional warnings](https://github.com/DiligentGraphics/DiligentCore/blob/master/CMakeLists.txt#L325).

PR #264, opened 2021-08-01
Spelling fixes and improved .gitignore

logankaser:
Now separated from the whitespace changes through the power of `git-apply --ignore-whitespace`


TheMostDiligent:
Could you please update this PR? There are some conflicts with the whitespace fix.

logankaser:
> Could you please update this PR? There are some conflicts with the whitespace fix.

Resolved now

TheMostDiligent:
Can you rebase this onto master? It can't be rebased automatically now.

logankaser:
Rebased

.gitignore
```diff
@@ -1,2 +1,31 @@
...
+.LSOverride
+xcuserdata/
+
+# Windows
+Thumbs.db
+Thumbs.db:encryptable
+ehthumbs.db
+ehthumbs_vista.db
+# Folder config file
+[Dd]esktop.ini
```
TheMostDiligent, line 31: Why do you have all these files in the root directory, BTW? Are you not using the build folder for your build files?

Graphics/GraphicsEngineVulkan/src/VulkanTypeConversions.cpp
```diff
@@ -1262,7 +1262,7 @@ static VkAccessFlags ResourceStateFlagToVkAccessFlags(RESOURCE_STATE StateFlag)
         case RESOURCE_STATE_BUILD_AS_READ:     return VK_ACCESS_SHADER_READ_BIT; // for vertex, index, transform, AABB, instance buffers
         case RESOURCE_STATE_BUILD_AS_WRITE:    return VK_ACCESS_ACCELERATION_STRUCTURE_READ_BIT_KHR | VK_ACCESS_ACCELERATION_STRUCTURE_WRITE_BIT_KHR; // for scratch buffer
         case RESOURCE_STATE_RAY_TRACING:       return VK_ACCESS_SHADER_READ_BIT; // for SBT
-        case RESOURCE_STATE_COMMON:            return 0; // COMMON state must be used for queue to queue transition (linke in D3D12), queue to queue synchronization via semaphore creates a memory dependency
+        case RESOURCE_STATE_COMMON:            return 0; // COMMON state must be used for queue to queue transition (linked in D3D12), queue to queue synchronization via semaphore creates a memory dependency
```
TheMostDiligent, line 1265: This was supposed to mean "similar to D3D12"

TheMostDiligent (changes requested):

.gitignore
logankaser: Most of these files are very common additions to git ignores because they are files created by the OS for varius reasons. I can remove the cmake files if you perfer.

For Linux the `.ntfs*` catches temporary files from network file shares.
Most of the macOS ones are caching of various folder information, with `.DS_Store` being very, very commonly created when interacting with a folder, and for Windows, the thumbnail files are mostly for older versions of Windows as far as I know, but I'm not that familiar with development on that platform

Graphics/GraphicsEngineVulkan/src/VulkanTypeConversions.cpp
logankaser: Ah, you meant "like" not "linked" I will fix.

TheMostDiligent: Yes

.gitignore
TheMostDiligent: Yes, I see. I was just trying to keep .gitignore as lean as possible, and on all platforms if you generate build files in build folder, all of them were ignored.
VSCode and Linux make sense to me.  For Windows and Mac, should these files not be in the build folder?

logankaser: `.DS_Store` for example will usually be generated if you were to example look at the repo root in Finder.
The reason is that the OS likes to cache folder attributes like icons and other stuff for easy access.
Same with windows for some versions if you open a folder with pictures in, as Explorer.exe generates thumbnails they are cached in a `Thumbs.db`. Neither has anything to do with build artifacts. `xcuserdata/` is arguably not needed, it is only likely to appear if you actually use Xcode as an IDE, most people I know on macOS only use it as a build tool, however.

If you would rather keep it lean, I can remove this stuff, it's just a minor development quality of life thing after all, does not affect using the library.

PR #265, opened 2021-08-01
Fix whitespace

logankaser:
Fix trailing whitespace for the entire repo (skipping markdown, of course)
Motivated by editors that strip whitespace causing large messy diffs


TheMostDiligent:
Why does this change add 70 lines if it is white space fix only?
Also, it breaks formatting.

logankaser:
> Why does this change add 70 lines if it is white space fix only?
> Also, it breaks formatting.

Not sure, I used `git diff -U0 -w --no-color | git apply --cached --ignore-whitespace --unidiff-zero -` to stage only the non-whitespace files, then staged these the remainder in a separate one.
I suspect the 70 lines are files that were missing a trailing newline.
I'll run formatting on it, I forgot since I'm sued to that running automatically.

logankaser:
I figured it out, since the git ignore changes got move to the other PR, some files from vscode got added to this PR, should be clean now.

Issue #208, opened 2021-04-17
Consider running codespell as part of CI


TheMostDiligent:
https://github.com/codespell-project/codespell

TheMostDiligent:
https://github.com/topics/proofreading

TheMostDiligent:
See #337

PR #268, opened 2021-08-16
Fix typos

luzpaz:
Found via `codespell -q 3 -S ./ThirdParty,*.gltf -L boundimg,delimeter,delimeters,inout,lod,pres`


TheMostDiligent:
Could you take a look at how to make codespell run up as part of the CI validations, e.g. [here](https://github.com/DiligentGraphics/DiligentCore/blob/master/.github/workflows/linux.yml#L20)?
It would be nice to catch all typos at merge stage.
https://github.com/DiligentGraphics/DiligentCore/issues/208

PR #273, opened 2021-09-07
check if Vulkan_LIBRARY defined, before using it

AndreyMlashkin:
I am building Vulkan libs from sources and I don't have dxcompiler lib and I also don't have Vulkan_LIBRARY  variable defined.
Therefore, I am getting an error get_filename_component got invalid number of arguments.

In any case, it's good to check if variable exists before actually using it

### DIFF
--- a/Graphics/GraphicsEngineVulkan/CMakeLists.txt
+++ b/Graphics/GraphicsEngineVulkan/CMakeLists.txt
@@ -176,7 +176,7 @@ endif()
 if(PLATFORM_WIN32)
     # find Vulkan SDK
     find_package(Vulkan)
-    if(Vulkan_FOUND)
+    if(Vulkan_FOUND AND DEFINED ${Vulkan_LIBRARY})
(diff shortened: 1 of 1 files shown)


TheMostDiligent:
Do you know why CMake gets into this branch? If you are building from source, then it should probably not get into this branch at all.

AndreyMlashkin:
Anyway, it is good to check if Vulkan_LIBRARY variable is defined right before using it ;)

TheMostDiligent:
This variable should be set by `find_package`, which also sets the `Vulkan_FOUND`. If the package is not found, `Vulkan_FOUND` should not be set and `Vulkan_LIBRARY` should not be checked.
I don't understand why `Vulkan_FOUND` is set, but `Vulkan_LIBRARY` is not.

TheMostDiligent:
After second thought, I am also confused which exactly vulkan libraries you are building from source. Diligent uses Volk as Vulkan loader. It does not use vulkan.lib

PR #278, opened 2021-09-20
Fix typos

tombish:
Discovered by CLion proofreading tool


TheMostDiligent:
Please use `-DDILIGENT_BUILD_TESTS=TRUE` to build the tests

Tests/DiligentCoreAPITest/src/ClearRenderTargetTest.cpp
```diff
@@ -121,7 +121,7 @@ TEST(ClearRenderTargetTest, AsRenderTarget)
     auto* pEnv       = TestingEnvironment::GetInstance();
     auto* pDevice    = pEnv->GetDevice();
     auto* pSwapChain = pEnv->GetSwapChain();
-    auto* pContext   = pEnv->GetDeviceContext();
```
TheMostDiligent: Here and below pContext is the same, so the second definition should simply be removed

tombish: :grimacing: Sorry made a mess of this should have seen that.

TheMostDiligent:
Linux build is failing because you fixed typo in `GeometryShaderRefenceGL` in CMake, but did not rename the file. Same with `TessellationRefenceGL`

tombish:
Would It be worthwhile starting a new pull request with all changes under one commit, to avoid the messy commit history?

PR #279, opened 2021-09-23
Minor changes for Emscripten

MikhailGorobets:
Minor changes in GitHub Actions and CMake files

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -17,14 +17,15 @@ jobs:
       with:
         submodules: recursive
 
-    - name: Download Ninja
-      uses: actions/checkout@master
-
     - name: Configure Ninja
       uses: seanmiddleditch/gha-setup-ninja@master
 
     - name: Download and Configure Emscripten SDK
-      uses: mymindstorm/setup-emsdk@v9
+      if: success()
+      shell: bash
+      run: |
+        cd $GITHUB_WORKSPACE/BuildTools/Scripts/github_actions
+        ./emscripten_sdk.sh "${{runner.workspace}}" 
 
     - name: Configure CMake
       if: success()
@@ -53,14 +54,16 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Download Ninja
-        uses: actions/checkout@master
-
       - name: Configure Ninja
         uses: seanmiddleditch/gha-setup-ninja@master
 
       - name: Download and Configure Emscripten SDK
-        uses: mymindstorm/setup-emsdk@v9
+        if: success()
+        shell: bash
+        run: |
+          cd $GITHUB_WORKSPACE/BuildTools/Scripts/github_actions
+          chmod +x emscripten_sdk.sh
+          ./emscripten_sdk.sh "${{runner.workspace}}" 
 
       - name: Configure CMake
         if: success()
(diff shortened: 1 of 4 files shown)


Graphics/HLSL2GLSLConverterLib/CMakeLists.txt
```diff
@@ -25,6 +25,20 @@ set(GLSL_DEFINITIONS include/GLSLDefinitions.h)
 set(GLSL_DEFINITIONS_INC ${CMAKE_CURRENT_SOURCE_DIR}/include/GLSLDefinitions_inc.h)
 set_source_files_properties(${GLSL_DEFINITIONS_INC} PROPERTIES GENERATED TRUE)
 
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
TheMostDiligent, line 28: I would recommend using `(void)partitioning` in the source code.

Graphics/GraphicsEngineOpenGL/CMakeLists.txt
```diff
@@ -231,6 +231,16 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
             COMPILE_FLAGS " -Wno-unused-variable"
         )
     endif()
+
```
TheMostDiligent, line 234: use `(void) GLFormat`

.github/workflows/emscripten.yml
```diff
@@ -24,7 +24,9 @@ jobs:
       uses: seanmiddleditch/gha-setup-ninja@master
 
     - name: Download and Configure Emscripten SDK
-      uses: mymindstorm/setup-emsdk@v9
+      uses: mymindstorm/setup-emsdk@master
+      with: 
+        version: 2.0.30
```
TheMostDiligent, line 29: Can we define this version so that it is shared between action files? We will have to add action files for 4 other submodules, it will be inconvenient to change version in 15 places

.github/workflows/emscripten.yml
```diff
@@ -1,5 +1,8 @@
 name: Emscripten
 
+env:
+  EMSCRIPTEN_SDK_VERSION: 2.0.30
```
TheMostDiligent, line 4: I would suggest importing it from a separate file that can be shared with other submodules.

MikhailGorobets: Yea. Perhaps it should be done by analogy with the file `configure_cmake.sh`

.github/workflows/emscripten.yml
```diff
@@ -23,8 +23,17 @@ jobs:
...
+    - name: Configure Emscripten SDK version
+      if: success()
+      shell: bash
+      run: |
+        cd $GITHUB_WORKSPACE/BuildTools/Scripts/github_actions
+        ./emscripten_sdk.sh
+
     - name: Download and Configure Emscripten SDK
-      uses: mymindstorm/setup-emsdk@v9
+      uses: mymindstorm/setup-emsdk@master
```
TheMostDiligent, line 34: Is it possible to run this command from `emscripten_sdk.sh`?

BuildTools/Scripts/github_actions/emscripten_sdk.sh
```diff
@@ -0,0 +1,2 @@
+EMSCRIPTEN_SDK_VERSION="2.0.30"
+echo "EMSCRIPTEN_SDK_VERSION=$EMSCRIPTEN_SDK_VERSION" >> $GITHUB_ENV
```
TheMostDiligent, line 2: New line at the end of the file is missing

.github/workflows/emscripten.yml
MikhailGorobets: I don't know. I'm experimenting

TheMostDiligent: The goal here is to maximize the reuse of logic. We would need to the same for other submodules, so it is preferable to have us little copy-paste as possible.

PR #296, opened 2021-11-06
Added Pipeline State Cache (API250012)

TheMostDiligent:


Graphics/GraphicsEngineVulkan/src/PipelineStateCacheVkImpl.cpp
```diff
@@ -0,0 +1,87 @@
...
+    }
+// clang-format on
+{
+    VkPipelineCacheCreateInfo VkPipelineStateCacheCI{};
+    VkPipelineStateCacheCI.sType = VK_STRUCTURE_TYPE_PIPELINE_CACHE_CREATE_INFO;
+
+    if (CreateInfo.pCacheData != nullptr && CreateInfo.CacheDataSize > 0)
+    {
+        VkPipelineStateCacheCI.initialDataSize = CreateInfo.CacheDataSize;
+        VkPipelineStateCacheCI.pInitialData    = CreateInfo.pCacheData;
```
MikhailGorobets, line 55: In theory, if the data isn't valid, then the cache data will be ignored.
But in practice, some (older) drivers neglect to verify the UUID correctly.
```
// Provided by VK_VERSION_1_0
typedef struct VkPipelineCacheHeaderVersionOne {
    uint32_t                        headerSize;
    VkPipelineCacheHeaderVersion    headerVersion;
    uint32_t                        vendorID;
    uint32_t                        deviceID;
    uint8_t                         pipelineCacheUUID[VK_UUID_SIZE];
} VkPipelineCacheHeaderVersionOne;
```

MikhailGorobets (approved):

Graphics/GraphicsEngineVulkan/src/PipelineStateCacheVkImpl.cpp
TheMostDiligent: Yes, I agree. We should check the driver/vendor version and reject the cache if there is a mismatch

azhirnov (approved):

PR #318, opened 2021-12-06
Graphics: Some updates and refactoring

MikhailGorobets:


Graphics/GraphicsEngine/interface/GraphicsTypes.h
```diff
@@ -1271,6 +1272,17 @@ struct DeviceObjectAttribs
...
+    /// Comparison operator tests if two structures are equivalent
+
+    /// \param [in] RHS - reference to the structure to perform comparison with
+    /// \return
+    /// - True if all members of the two structures are equal.
+    /// - False otherwise.
+    bool operator==(const DeviceObjectAttribs& RHS) const
+    {
+        return SafeStrEqual(Name, RHS.Name);
+    }
```
TheMostDiligent, line 1285: Objects with different names are considered the same if other fields are the same. E.g. two rasterizer states "Rasterizer State 1" and "Rasterizer State 2" are still the same if other fields are identical.
Several internal caching systems rely on that and this may break them.

Graphics/GraphicsEngine/interface/DepthStencilState.h
```diff
@@ -132,16 +132,16 @@ struct StencilOpDesc
...
-               StencilDepthFailOp == rhs.StencilDepthFailOp &&
-               StencilPassOp      == rhs.StencilPassOp      &&
-               StencilFunc        == rhs.StencilFunc;
+        return StencilFailOp      == RHS.StencilFailOp      &&
+               StencilDepthFailOp == RHS.StencilDepthFailOp &&
+               StencilPassOp      == RHS.StencilPassOp      &&
+               StencilFunc        == RHS.StencilFunc;
     }
 #endif
 };
```
TheMostDiligent, line 147: Here, and in other similar places, rhs is fine. No need to rename it.

TheMostDiligent (changes requested):

PR #327, opened 2021-12-21
Updated File2String (close #94)

### LINKED ISSUE #94: NVIDIA Jetson TX2 - Fix File2String errors
DiligentCore/Utilities/File2Include

Under the bin there is a File2String precompiled for each platform , but for Linux its compiled for x86 architecture and its not working on ARM .

There also problem with c++11 that should be defined to properly compile auto and nullptr

Probably this utility should be compiled first on the target system

MikhailGorobets:


TheMostDiligent:
Why is Emscripten build failing?

TheMostDiligent (changes requested):
Please fix Emscripten build

PR #340, opened 2022-01-25
Add callback wrapper

MikhailGorobets:


Common/interface/CallbackWrapper.hpp
```diff
@@ -0,0 +1,67 @@
...
+struct Callback
+{
+    void (*Function)(Arg, void*);
+    Lambda State;
+};
+
+template <typename Lambda, typename Arg = typename FunctionTraits<Lambda>::template Argument<0>>
+Callback<Lambda, Arg> WrapCallback(Lambda&& L)
+{
+    static_assert(FunctionTraits<Lambda>::Arity == 1, "Check number of arguments");
```
TheMostDiligent, line 57: There is a bug here: when using this as LoadPipelineStateInfo::Modify, the Arity is 2

Common/interface/CallbackWrapper.hpp
```diff
@@ -0,0 +1,67 @@
...
+    using Func = typename std::decay<Lambda>::type;
+    return {
+        [](Arg CI, void* pUserData) -> void {
+            Func* pFunc = static_cast<Func*>(pUserData);
+            (*pFunc)(CI);
```
TheMostDiligent, line 62: (*pFunc)(CI, pUserData)

Common/interface/CallbackWrapper.hpp
```diff
@@ -0,0 +1,67 @@
...
+        },
+        std::forward<Lambda>(L)};
+}
```
TheMostDiligent, line 65: Overall, this is a very narrow implementation for a specific case: the callback may only handle functions of type (Type, void*).
Try to think if it is possible to make wrapper for functions of any type.

TheMostDiligent (changes requested):

PR #344, opened 2022-02-08
Added Shader Bytecode Cache (close #339)

### LINKED ISSUE #339: Add Shader Bytecode Cache
Add Shader Bytecode Cache component to [GraphicsTools](https://github.com/DiligentGraphics/DiligentCore/tree/master/Graphics/GraphicsTools) project.

Usage scenario: an application creates the byte code cache at startup and loads the data from the disk (if any). During run-time the application uses the cache to get existing byte code and add new byte code it creates. At the end, the application writes the cache back to disk.

Possible API:

```cpp
class IBytecodeCache
{
    /// Loads the byte code data from the data blob
    ///
    /// \remarks    The method may be called multiple times with different
    ///             data chunks. The data from every chunk will be added to the
    ///             cache.
    virtual void Load(IDataBlob* pData) = 0;
 
    /// Tries to find the byte code in the cache. If byte code is found, creates the data blob
    /// that contains it. If bytecode is not found, *ppByteCode will be null.
 
    // Implementation suggestions:
    //  - The method should construct the hash of all relevant data in the ShaderCI
    //    - Source, macros, compiler flags, shader version etc.
    //  - For includes, it should recursively process all includes
    //    - Implement a basic algorithm that will find all #include directives not
    //      in comments
    //  - Use hash to find the byte code
    //    - If StoreSource is true, compare the source code matches the data in the cache
    //  - Use custom implementation of IDataBlob that will not allocate memory but
    //    instead will be a view of the relevant data.
    virtual void GetBytecode(const ShaderCreateInfo& ShaderCI,
                             IDataBlob**             ppByteCode) = 0;
 
    /// Adds shader byte code to the cache.
    virtual void AddBytecode(const ShaderCreateInfo& ShaderCI,
                             IDataBlob*              pByteCode) = 0;
 
    /// Removes shader byte code from the cache.
    virtual void RemoveBytecode(const ShaderCreateInfo& ShaderCI) = 0;
 
    /// Stores the cache data in the data blob.
    virtual void Store(IDataBlob* pDataBlob) = 0;
 
    /// Clears the cache.
    virtual void Clear() = 0;
};
 
struct BytecodeCacheCreateInfo
{
    /// Shader source language to use when ShaderCI.SourceLanguage is SHADER_SOURCE_LANGUAGE_DEFAULT.
    SHADER_SOURCE_LANGUAGE DefaultLanguage DEFAULT_INITIALIZER(SHADER_SOURCE_LANGUAGE_DEFAULT);
 
    /// Whether to store original source in the cache
    bool StoreSource DEFAULT_INITIALIZER(false);
};
 
void DILIGENT_GLOBAL_FUNCTION(CreateBytecodeCache)(const BytecodeCacheCreateInfo& CreateInfo,
                                                   IBytecodeCache**               ppLoader);

```

- [x] Add shader cache to the Diligent-GraphicsTools project in the core module
- [x] Add unit tests
- [x] It may be worth adding the GetHash() function to the common shader API to facilitate shader caching. It may be useful later.

MikhailGorobets:


Graphics/GraphicsTools/src/BytecodeCache.cpp
```diff
@@ -0,0 +1,330 @@
...
+#include "ObjectBase.hpp"
+
+#include "BytecodeCache.h"
+#include "ShaderPreprocessor.hpp"
+#include "XXH128Hasher.hpp"
+
+namespace
+{
+
+class InputMemoryStream
```
TheMostDiligent, line 40: I suggest moving InputMemoryStream and OutputMemoryStream to Diligent-Primitives (e.g. MemoryStream.hpp), and having dedicated tests for them.

Graphics/GraphicsTools/src/BytecodeCache.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+        BytecodeCacheHeader Header{};
+        Stream.Read(Header);
+
+        for (Uint64 ItemID = 0; ItemID < Header.CountElements; ItemID++)
+        {
+            BytecodeCacheElementHeader ElementHeader{};
+            Stream.Read(ElementHeader);
+
+            auto pBytecode = DataBlobImpl::Create(static_cast<size_t>(ElementHeader.DataSize));
```
TheMostDiligent, line 188: `StaticCast<>` to be safe

Graphics/GraphicsTools/src/XXH128Hasher.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include "xxhash.h"
+
+#include "XXH128Hasher.hpp"
+#include "DebugUtilities.hpp"
+
+namespace Diligent
+{
+
+XXH128State::XXH128State() :
+    m_State(XXH3_createState())
```
TheMostDiligent, line 36: Brace initialization

TheMostDiligent (changes requested):

Graphics/GraphicsTools/src/BytecodeCache.cpp
```diff
@@ -0,0 +1,330 @@
...
+        Stream.Write(ShaderCI.UseCombinedTextureSamplers);
+        Stream.Write(ShaderCI.Desc.ShaderType);
+        Stream.Write(ShaderCI.ShaderCompiler);
+        Stream.Write(ShaderCI.HLSLVersion);
+        Stream.Write(ShaderCI.GLSLVersion);
+        Stream.Write(ShaderCI.GLESSLVersion);
+        Stream.Write(ShaderCI.CompileFlags);
+        Stream.Write(m_DefaultLanguage);
+        Stream.Write(m_DeviceType);
+
```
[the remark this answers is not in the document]
MikhailGorobets: If we leave the code as it is, then we will not need to call strlen twice

Graphics/GraphicsTools/src/ShaderPreprocessor.cpp
```diff
@@ -0,0 +1,207 @@
...
+                        PreprocessorState = None; // Buggy code, skip over this include.
+                        break;
+                    case '\"':
+                        IncludeHandler(std::string(&pBuffer[Start], &pBuffer[Offset]));
+                        PreprocessorState = None;
+                        break;
+                }
+                break;
+        }
+        Offset++;
```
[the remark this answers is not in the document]
TheMostDiligent: What about the comments for the parser [as in the example](https://github.com/DiligentGraphics/DiligentCore/blob/master/Graphics/HLSL2GLSLConverterLib/src/HLSL2GLSLConverterImpl.cpp#L2495)

Graphics/GraphicsTools/src/BytecodeCache.cpp
```diff
@@ -0,0 +1,221 @@
...
+        BytecodeCacheHeader Header{};
+        Header.CountElements = m_HashMap.size();
+        Stream.Write(Header);
+
+        for (auto const& Pair : m_HashMap)
+        {
+            Stream.Write(BytecodeCacheElementHeader{Pair.first, Pair.second->GetSize()});
+            Stream.Write(Pair.second->GetConstDataPtr(), Pair.second->GetSize());
+        }
```
TheMostDiligent, line 129: I suggest using Serializer here. This way you will not need a new OutputStream class

Graphics/GraphicsTools/src/ShaderPreprocessor.cpp
```diff
@@ -0,0 +1,198 @@
...
+                {
+                    case '\n':
+                        PreprocessorState = None; // Buggy code, skip over this include.
+                        break;
+                    case '\"':
+                        IncludeHandler(std::string(&pBuffer[Start], &pBuffer[Offset]));
+                        PreprocessorState = None;
+                        break;
+                }
+                break;
```
TheMostDiligent, line 136: Please add comments for parser operation - they will be very helpful to show what the parser is doing

Primitives/interface/MemoryStream.hpp
```diff
@@ -0,0 +1,135 @@
...
+ */
+
+#pragma once
+
+#include "BasicTypes.h"
+
+namespace Diligent
+{
+
+class InputMemoryStream
```
TheMostDiligent, line 34: Please add some tests for the object

TheMostDiligent (changes requested):

PR #348, opened 2022-02-16
Added shader include inlining function (close #346)

### LINKED ISSUE #346: Add shader include inlining function
Add a function that will inline all shader includes in a source file, removing original `#include` directives so that the source can be compiled.

MikhailGorobets:


Graphics/ShaderTools/include/ShaderToolsCommon.hpp
```diff
@@ -80,15 +80,24 @@ const char* ReadShaderSourceFile(const char*                      SourceCode,
...
+    /// The function to let the application to get data of the included file
+    std::function<void(IDataBlob* pDataBlob, const Char* FilePath)> IncludeHandler{};
+
+    /// The function to let the application to modify #include "${PATH}" substring in the source file
+    ///
+    ///      #include "SomeInclude.hlsl"
+    ///      ^                         ^
+    ///    Start                      End
+    std::function<void(Char* Start, Char* End)> IncludeModify{};
 };
```
TheMostDiligent, line 94: I think it is better to add `Char* Start; Char* End;` to `ProcessShaderIncludesInfo` struct so that there is only one callback function that can do what it needs with data and modify the string at the same time.

It is also not clear if IncludeHandler returns data before modification or after - this will be clear if it was one function.

Also I think that End is the position *past* the last symbol, not the last symbol itself

```
    ///      #include "SomeInclude.hlsl"
    ///      ^                          ^
    ///    Start                      End
```

Graphics/ShaderTools/src/ShaderToolsCommon.cpp
```diff
@@ -327,4 +331,32 @@ bool ProcessShaderIncludes(const ShaderCreateInfo& ShaderCI, std::function<void(
...
 
+void MergeShaderIncludes(const ShaderCreateInfo& ShaderCI, IDataBlob** ppData)
+{
+    VERIFY_EXPR(ShaderCI.Desc.Name != nullptr);
+    VERIFY_EXPR(ppData != nullptr);
+
+    std::stringstream Stream;
+
+    auto IncludeHandler = [&](IDataBlob* pData, const Char* Path) {
+        Stream << String{static_cast<const Char*>(pData->GetConstDataPtr()), pData->GetSize()} << std::endl;
```
TheMostDiligent, line 342: I think you can use `Stream.write` here to avoid creating temporary string and allocating memory that will be released immediately after

Graphics/ShaderTools/src/ShaderToolsCommon.cpp
```diff
@@ -327,4 +331,32 @@ bool ProcessShaderIncludes(const ShaderCreateInfo& ShaderCI, std::function<void(
...
+
+    IncludePreprocessInfo Info{};
+    if (!ProcessShaderIncludes(ShaderCI, {IncludeHandler, IncludeModify}))
+    {
+        LOG_ERROR("Failed to merge includes for shader: '", ShaderCI.Desc.Name, "'.");
+        return;
+    }
+
+    const auto InlineData = Stream.str();
+    auto       pData      = DataBlobImpl::Create(InlineData.size(), InlineData.data());
```
TheMostDiligent, line 357: There is `StringDataBlobImpl` that can assume ownership of the `std::string` without making copies

Tests/DiligentCoreTest/src/ShaderTools/ShaderPreprocessTest.cpp
```diff
@@ -123,13 +123,43 @@ TEST(ShaderPreprocessTest, Include)
...
+    ASSERT_NE(pShaderSourceFactory, nullptr);
+
+    {
+        ShaderCreateInfo ShaderCI{};
+        ShaderCI.Desc.Name                  = "TestShader";
+        ShaderCI.FilePath                   = "InlineIncludeShaderTest.hlsl";
+        ShaderCI.pShaderSourceStreamFactory = pShaderSourceFactory;
+
+        const std::string Reference =
+            "// InlineIncludeShaderCommon0.hlsl\n\n"
```
TheMostDiligent, line 148: Can you add include in the comment to the test, e.g.
/* #include "Non-existing" */
Also, please add `#include <>`

TheMostDiligent (changes requested):

.gitattributes
```diff
@@ -0,0 +1 @@
+* text eol=lf
```
TheMostDiligent, line 1: Why is this needed? Most of the files use CRLF

Graphics/ShaderTools/include/ShaderToolsCommon.hpp
```diff
@@ -81,14 +81,28 @@ const char* ReadShaderSourceFile(const char*                      SourceCode,
...
+    /// The path to the included file.
+    const Char* FilePath;
+
+    /// Offsets relative to the pDataBlob->GetDataPtr() pointer to modify a #include "${PATH}" substring
+    ///
+    ///      #include "SomeInclude.hlsl"
+    ///      ^                         ^
+    ///    Start                      End
+    size_t Start;
+    size_t End;
```
TheMostDiligent, line 98: Always use default initializers.

Graphics/ShaderTools/src/ShaderToolsCommon.cpp
```diff
@@ -173,47 +175,48 @@ static void ExtractDependencies(std::function<void(const Diligent::String&)> Inc
...
-    //We iterate over the characters of the buffer
+    /// We iterate over the characters of the buffer
     while (Offset < BufferSize)
     {
         switch (PreprocessorState)
         {
             case None:
             {
-                // Find a hash character if the current position is greater NextHash
+                /// Find a hash character if the current position is greater NextHash
```
TheMostDiligent, line 195: `///` Is for doxygen. These comment are implementation details. They don't need to be in the documentation.

Tests/DiligentCoreTest/src/ShaderTools/ShaderPreprocessTest.cpp
```diff
@@ -123,13 +123,43 @@ TEST(ShaderPreprocessTest, Include)
...
+            "// #include \"InlineIncludeShaderTest.hlsl\"\n\n";
+
+        RefCntAutoPtr<IDataBlob> pData;
+        MergeShaderIncludes(ShaderCI, &pData);
+        ASSERT_NE(pData, nullptr);
+
+        const std::string DataString = std::string{static_cast<const char*>(pData->GetConstDataPtr()), pData->GetSize()};
+
+        ASSERT_EQ(Reference.size(), DataString.size());
+        EXPECT_EQ(Reference, DataString);
```
TheMostDiligent, line 161: There is `EPXECT_STREQ` macros for comparing c strings

TheMostDiligent (approved):

PR #350, opened 2022-02-17
FileSystem: Implemented some functions

MikhailGorobets:


Platforms/Apple/src/AppleFileSystem.cpp
```diff
@@ -125,14 +127,28 @@ bool AppleFileSystem::FileExists(const Char* strFilePath)
...
 }
 
 bool AppleFileSystem::CreateDirectory(const Char* strPath)
 {
-    UNSUPPORTED("Not implemented");
-    return false;
+    std::string path(strPath);
+    CorrectSlashes(path, AppleFileSystem::GetSlashSymbol());
+
+    auto res = mkdir(path.c_str(), 0777);
```
TheMostDiligent, line 142: Please use named constants instead of 777.
This should be S_IRWXU | S_IRWXG | S_IRWXO I think - double check

Platforms/Apple/src/AppleFileSystem.cpp
```diff
@@ -125,14 +127,28 @@ bool AppleFileSystem::FileExists(const Char* strFilePath)
...
+    return res == 0;
+}
+
+void AppleFileSystem::SetWorkingDirectory(const Char* strWorkingDir)
```
TheMostDiligent, line 146: Either remove this or add to Linux platform

TheMostDiligent (changes requested):

PR #372, opened 2022-05-08
Use copied GeometryName when creating BLAS instances.

threax:
Use the copied name in pTriangles or pBoxes in the BottomLevelASDesc. This allows these values to come from C#. When used from there with P/Invoke the SrcGeoName string is controlled by the marshaller and will be deleted once the BLAS has been created. Later, when we go to use the BLAS, the name will have an invalid value. By using the copy that is already created we avoid this situation and don't need further complex management to keep these strings initialized.

I've been using this for a while on 2.4.g and now 2.5.2 and it does not have any negative impacts that I can see.

### DIFF
--- a/Graphics/GraphicsEngine/src/BottomLevelASBase.cpp
+++ b/Graphics/GraphicsEngine/src/BottomLevelASBase.cpp
@@ -126,7 +126,8 @@ void CopyBLASGeometryDesc(const BottomLevelASDesc& SrcDesc,
         for (Uint32 i = 0; i < SrcDesc.TriangleCount; ++i)
         {
             const auto* SrcGeoName     = SrcDesc.pTriangles[i].GeometryName;
-            pTriangles[i].GeometryName = MemPool.CopyString(SrcGeoName);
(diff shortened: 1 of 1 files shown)


Graphics/GraphicsEngine/src/BottomLevelASBase.cpp
```diff
@@ -136,7 +136,7 @@ void CopyBLASGeometryDesc(const BottomLevelASDesc& SrcDesc,
                 ActualIndex = iter->second.ActualIndex;
             }
```
TheMostDiligent, line 137: I would recommend making the copy right away and always using it:
```cpp
const auto* GeoName    = MemPool.CopyString(SrcDesc.pBoxes[i].GeometryName);
pBoxes[i].GeometryName = GeoName;
```

TheMostDiligent (changes requested):

TheMostDiligent:
This is definitely a bug - thanks for finding it

TheMostDiligent:
Merged in https://github.com/DiligentGraphics/DiligentCore/commit/262341510cf6d5d75af83ee50e100a7645c7a6a0

PR #379, opened 2022-07-06
🐛: removed noexcept to pass exception on in case of OOM

ulifei:
When running out of GPU memory VulkanLogicalDevice::AllocateDeviceMemory throws an exception, however the calling function (VulkanMemoryPage::VulkanMemoryPage) was noexcept. This caused the application to terminate, instead of handling the error in RenderDeviceBase::CreateDeviceObject

### DIFF
--- a/Graphics/GraphicsEngineVulkan/include/VulkanUtilities/VulkanMemoryManager.hpp
+++ b/Graphics/GraphicsEngineVulkan/include/VulkanUtilities/VulkanMemoryManager.hpp
@@ -103,7 +103,7 @@ public:
     ~VulkanMemoryPage();
 
     // clang-format off
-    VulkanMemoryPage(VulkanMemoryPage&& rhs) :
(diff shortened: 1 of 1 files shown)


Graphics/GraphicsEngineVulkan/include/VulkanUtilities/VulkanMemoryManager.hpp
```diff
@@ -99,11 +99,11 @@ class VulkanMemoryPage
                      VkDeviceSize          PageSize,
                      uint32_t              MemoryTypeIndex,
                      bool                  IsHostVisible,
-                     VkMemoryAllocateFlags AllocateFlags) noexcept;
+                     VkMemoryAllocateFlags AllocateFlags);
     ~VulkanMemoryPage();
 
     // clang-format off
-    VulkanMemoryPage(VulkanMemoryPage&& rhs)noexcept :
```
TheMostDiligent, line 106: FYI, move constructors should always be `noexcept`.

PR #390, opened 2022-10-26
Fixed atomic type

MikhailGorobets:


TheMostDiligent:
There is no reason/benefit to replacing `ReferenceCounterValueType` with `Int32`. Just doing
`using ReferenceCounterValueType = Int32` will have the same effect and leave room for flexibility, if needed.

PR #405, opened 2023-05-06
Fix crash in LoadShaderCodeVariableDesc.

eugeneko:
Should be self-explanatory. Cannot pass `nullptr` to `std::string`


TheMostDiligent:
Fixed in https://github.com/DiligentGraphics/DiligentCore/commit/54760f1fe54e72f267c5fb83fb70ed747a66895a

PR #406, opened 2023-05-16
Fix bind target in BufferGLImpl::UpdateData.

eugeneko:
Using `GL_ARRAY_BUFFER` for index buffer causes this error on WebGL:

> bindBuffer: element array buffers can not be bound to a different target

I see no reason _not_ to use actual buffer type in UpdateData.

### DIFF
--- a/Graphics/GraphicsEngineOpenGL/src/BufferGLImpl.cpp
+++ b/Graphics/GraphicsEngineOpenGL/src/BufferGLImpl.cpp
@@ -224,13 +224,13 @@ void BufferGLImpl::UpdateData(GLContextState& CtxState, Uint64 Offset, Uint64 Si
         CtxState);
 
     // We must unbind VAO because otherwise we will break the bindings
-    constexpr bool ResetVAO = true;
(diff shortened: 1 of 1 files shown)


Graphics/GraphicsEngineOpenGL/src/BufferGLImpl.cpp
```diff
@@ -225,12 +225,12 @@ void BufferGLImpl::UpdateData(GLContextState& CtxState, Uint64 Offset, Uint64 Si
 
     // We must unbind VAO because otherwise we will break the bindings
     constexpr bool ResetVAO = true;
```
TheMostDiligent, line 227: I believe VAO needs to be reset only for `ARRAY_BUFFER` and `ELEMENT_ARRAY_BUFFER`:
```cpp
const auto ResetVAO = m_BindTarget == GL_ARRAY_BUFFER || m_BindTarget == GL_ELEMENT_ARRAY_BUFFER;
```

TheMostDiligent:
Can you rebase the commit - I can't do it from the UI

eugeneko:
Done. Also squashed commits into one because I can.

PR #413, opened 2023-06-21
Support read-only DSV in ITextureView.

eugeneko:
Per discussion in Discord.


eugeneko:
It is ready now.

I have written RenderPass-based test in addition to RenderTarget-based test, but I didn't commit it because I don't think it's useful.
Render passes already have all resource transitions explicitly described, so `TEXTURE_VIEW_READ_ONLY_DEPTH_STENCIL` has no utility when using render passes.

eugeneko:
PS: I've just noticed that my IDE automatically removes trailing spaces. I can revert those if requested.

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
```diff
@@ -0,0 +1,614 @@
...
+        m_pReadWriteDSV.Release();
+        m_pReadOnlyDSV.Release();
+
+        m_pRT.Release();
+        m_pRTV.Release();
+
+        m_pColorSRB.Release();
+    }
+
+    void SkipIfNotSupported()
```
MikhailGorobets, line 125: GTEST_SKIP_ cannot be called outside of a test function. Because of this the test with format VK_FORMAT_D24_UNORM_S8_UINT does not skip on AMD graphics cards

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
```diff
@@ -0,0 +1,614 @@
...
+    {
+        auto* pEnv    = GPUTestingEnvironment::GetInstance();
+        auto* pDevice = pEnv->GetDevice();
+
+        const bool bSupported = !!(pDevice->GetTextureFormatInfoExt(GetDepthFormat()).BindFlags & BIND_DEPTH_STENCIL);
```
MikhailGorobets, line 130: Why do you need a double negation operation?

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
```diff
@@ -0,0 +1,614 @@
...
+using namespace Diligent;
+using namespace Diligent::Testing;
+
+namespace
+{
+
+// clang-format off
+const std::string ReadOnlyDepthTest_VS{
+R"(
+struct PSInput
```
MikhailGorobets, line 42: This shader can be shortened by passing the arguments directly to the function

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
```diff
@@ -0,0 +1,614 @@
+/*
+ *  Copyright 2019-2022 Diligent Graphics LLC
+ *  Copyright 2015-2019 Egor Yusov
```
MikhailGorobets, line 3: Change to `Copyright 2019-2023 Diligent Graphics LLC`

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
```diff
@@ -0,0 +1,614 @@
...
+{
+    float depth = g_Input.Load(int3(0, 0, 0)).r;
+    return float4(depth, depth * 0.5, 0.75, 1.0);
+}
+)"
+};
+// clang-format on
+
+// Arbitrary clear color and depth that are never used anywhere else in this test
+constexpr float ClearColor[] = {0.11f, 0.32f, 0.02f, 1.0f};
```
MikhailGorobets, line 80: Use floating point numbers such as  `0.125, 0.250, 0.5` etc

Graphics/GraphicsEngine/interface/PipelineState.h
```diff
@@ -340,6 +344,7 @@ struct GraphicsPipelineDesc
               SubpassIndex      == Rhs.SubpassIndex      &&
               ShadingRateFlags  == Rhs.ShadingRateFlags  &&
               DSVFormat         == Rhs.DSVFormat         &&
+              UseReadOnlyDSV    == Rhs.UseReadOnlyDSV    &&
```
MikhailGorobets, line 347: Still, I don't like the idea of additional fields. Can we find a value for the render pass based on the DepthStencilState?
@TheMostDiligent

Graphics/GraphicsAccessories/src/GraphicsAccessories.cpp
```diff
@@ -81,12 +81,14 @@ class TexFormatToViewFormatConverter
         static_assert(TEXTURE_VIEW_RENDER_TARGET == 2, "TEXTURE_VIEW_RENDER_TARGET == 2 expected");
         static_assert(TEXTURE_VIEW_DEPTH_STENCIL == 3, "TEXTURE_VIEW_DEPTH_STENCIL == 3 expected");
         static_assert(TEXTURE_VIEW_UNORDERED_ACCESS == 4, "TEXTURE_VIEW_UNORDERED_ACCESS == 4 expected");
+        static_assert(TEXTURE_VIEW_READ_ONLY_DEPTH_STENCIL == 6, "TEXTURE_VIEW_READ_ONLY_DEPTH_STENCIL == 6 expected");
```
MikhailGorobets, line 84: Maybe we should add four flags at once? @TheMostDiligent 
```
Depth: Read, Stencil: Read
Depth: Read, Stencil: Write
Depth: Write, Stencil: Read
Depth: Write, Stencil: Write 
```

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
eugeneko: This was very confusing behavior of GTest. I expected `GTEST_SKIP_` to behave like asserts. Fixed.

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
eugeneko: To cast integer to bool without any warnings. I see that Diligent usually does (a & b) != 0 instead, I will change it for consistency.

Graphics/GraphicsEngine/interface/PipelineState.h
eugeneko: We _can_ do that, but it will (1) silently break existing user code, and (2) will require extra render target changes in user code. Which is the trade off I am personally not willing to take.

Graphics/GraphicsAccessories/src/GraphicsAccessories.cpp
eugeneko: `TEXTURE_VIEW_DEPTH_STENCIL` directly corresponds to `RESOURCE_STATE_DEPTH_WRITE`.
`TEXTURE_VIEW_READ_ONLY_DEPTH_STENCIL` directly corresponds to `RESOURCE_STATE_DEPTH_READ`.

What would "Depth: Read, Stencil: Write" and "Depth: Write, Stencil: Read" correspond to?

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
eugeneko: Fixed. I forgot about that.

Graphics/GraphicsAccessories/src/GraphicsAccessories.cpp
MikhailGorobets: In Vulkan there are such flags, in d3d12 respectively there is a way to create such combinations
```
 VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL
 VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL
 VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL
 VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL
 ```

Comment (eugeneko) [reply Graphics/GraphicsAccessories/src/GraphicsAccessories.cpp:84]: Oh, I didn't know that existed.
I agree that having those may be helpful in general.
However, adding those would require extending `RESOURCE_STATE`, and refactoring that I am afraid I am not quite qualified to do.

If you people decide you want to have all 4 combinations, I would suggest to extract it into separate task, and I will tweak this PR to minimize future changes.

Comment (MikhailGorobets) [reply Graphics/GraphicsAccessories/src/GraphicsAccessories.cpp:84]: Let's wait and see what Egor says. Perhaps it would be better to switch to the option with four states and also consider the possibility of adding dynamic rendering, so that we don't have to add additional fields to the Pipeline State.

File: Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp [hunk 8]
```diff
@@ -0,0 +1,581 @@
...
+
+        // clang-format off
+        LayoutElement Elems[] =
+        {
+            LayoutElement{ 0, 0, 4, VT_FLOAT32}
+        };
+        // clang-format on
+
+        GraphicsPipeline.InputLayout.LayoutElements = Elems;
+        GraphicsPipeline.InputLayout.NumElements    = 1;
```
MikhailGorobets, line 218: You can use `_countof` macro

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
```diff
@@ -0,0 +1,581 @@
...
+
+    pContext->CommitShaderResources(m_pColorSRB, RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
+
+    // Draw color fullscreen quad that reads depth from the texture and performs depth test simultaneously
+    ClearValues[0].Format = RTVFormat;
+    memcpy(ClearValues[0].Color, ClearColor, sizeof(ClearColor));
+
+    RPBeginAttribs.pRenderPass         = pColorRenderPass;
+    RPBeginAttribs.pFramebuffer        = pColorFramebuffer;
+    RPBeginAttribs.ClearValueCount     = 2;
```
MikhailGorobets, line 549: You can use `_countof` macro

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
```diff
@@ -0,0 +1,581 @@
...
+    pContext->SetRenderTargets(1, &m_pRTV, m_pReadWriteDSV, RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
+    pContext->ClearRenderTarget(m_pRTV, ClearColor, RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
+
+    CLEAR_DEPTH_STENCIL_FLAGS ClearFlags = CLEAR_DEPTH_FLAG;
+    if (GetDepthFormat() == TEX_FORMAT_D24_UNORM_S8_UINT || GetDepthFormat() == TEX_FORMAT_D32_FLOAT_S8X24_UINT)
+        ClearFlags |= CLEAR_STENCIL_FLAG;
+    pContext->ClearDepthStencil(m_pReadWriteDSV, ClearFlags, ClearDepth, 0, RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
+
+    OptimizedClearValue    ClearValues[2];
+    BeginRenderPassAttribs RPBeginAttribs;
```
MikhailGorobets, line 522: I think it's better not to reuse the same structure. It's preferable to create two variables and separate the code using a scope.

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
```diff
@@ -0,0 +1,581 @@
...
+            float4(-1.0f,  -1.0f,  GeometryDepth,  1.f),
+            float4(+1.0f,  +1.0f,  GeometryDepth,  1.f),
+            float4(+1.0f,  -1.0f,  GeometryDepth,  1.f)
+        };
+        // clang-format on
+
+        BufferDesc BuffDesc;
+        BuffDesc.Name      = "Test vertex buffer";
+        BuffDesc.BindFlags = BIND_VERTEX_BUFFER;
+        BuffDesc.Size      = sizeof(VertexData);
```
MikhailGorobets, line 395: For your case, I don't think a vertex buffer is necessary. You can generate it directly in a vertex shader.

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
```diff
@@ -0,0 +1,546 @@
...
+
+    InitializePipelineStates(pDepthRenderPass, pColorRenderPass);
+    InitializeSRB();
+
+    // Clear color and depth to unused colors
+    pContext->SetRenderTargets(1, &m_pRTV, m_pReadWriteDSV, RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
+    pContext->ClearRenderTarget(m_pRTV, ClearColor, RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
+
+    CLEAR_DEPTH_STENCIL_FLAGS ClearFlags = CLEAR_DEPTH_FLAG;
+    if (GetDepthFormat() == TEX_FORMAT_D24_UNORM_S8_UINT || GetDepthFormat() == TEX_FORMAT_D32_FLOAT_S8X24_UINT)
```
MikhailGorobets, line 475: Check above

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
```diff
@@ -0,0 +1,546 @@
...
+)"
+};
+// clang-format on
+
+// Arbitrary clear color and depth that are never used anywhere else in this test
+constexpr float ClearColor[] = {0.125f, 0.250f, 0.5f, 1.0f};
+constexpr float ClearDepth   = 0.125f;
+
+constexpr float ExpectedDepth    = 0.5f;
+constexpr float ReferenceColor[] = {ExpectedDepth, ExpectedDepth * 0.5f, 0.75f, 1.0f};
```
MikhailGorobets, line 86: `ExpectedDepth` became unnecessary, you deleted a vertex buffer

Tests/DiligentCoreAPITest/src/ReadOnlyDepthTest.cpp
eugeneko: Sorry, I don't follow. Could you elaborate, please?

MikhailGorobets: Yes you fixed it with `IsDepthStencilFormat`. 
Wouldn't it be better to create a function that directly returns CLEAR_DEPTH_STENCIL_FLAGS based on a format?

Graphics/GraphicsAccessories/src/GraphicsAccessories.cpp
TheMostDiligent: Yes, I think adding four flags makes sense for completeness.
Dynamic rendering would be and ideal solution here, only that this is a very big change.

Graphics/GraphicsEngine/interface/PipelineState.h
TheMostDiligent: > Still, I don't like the idea of additional fields.

Neither do I.

>  Can we find a value for a render pass based on DepthStencilState?

Probably. We can check the `DepthWriteEnable` flag and create the read-only depth. However this will create a problem with binding normal DSV and using it with the read-only depth pipeline.
I was thinking to have two views inside each DSV - one normal, one readonly, but then again there is a problem with mixing normal and read-only pipelines.

In current approach, what happens if you bind read-only depth pipeline, set render targets and then use read-write depth pipeline?

We have the `SetRenderTargetsExt` call. It looks like we will need to add ReadOnlyDepth flag to the struct it takes.

eugeneko: @TheMostDiligent 
Sorry, I don’t follow your question. The act of binding pipeline and setting render targets are still independent from each other. It’s okay if you have mismatched pipeline bound at some point as long as you don’t use it.

This is the reason why I don’t want to make automatic flagless detection — it’s gonna mess this up. We will have to have two pipelines in pipeline object if depth write is disabled. One if user decides to use PSO with read write depth and another if with read only. And this will also create non trivial engagement between setting PSO and render target. I would even say that effective PSO will have to be resolved in prepareDraw.

PR #418, opened 2023-07-21
Add callback for OpenGL program linkage.

eugeneko:
**Disclaimer:** I know that this code is an abomination. However, (1) it has functional impact that was crucial for me, (2) I would like to have this code visible and available for other people with similar issue.

If this issue is not resolved in the master branch of Diligent one way or another, my solution will be available in [this branch](https://github.com/rbfx/DiligentCore/tree/gl-hook) indefinetely for anyone to cherry-pick.

---

### My use case

- I have "dynamic" shaders which are compiled on demand in real time, instead of being generated and compiled in build time. On one hand, this is limitation of the legacy codebase. On the other hand, it enables some rare use cases, e.g. ShaderToy-like sandbox. Point is, *I must be able to compile shaders in runtime*.
- Shaders are authored in GLSL.
- For OpenGL backend, using glslang/SPIRV-Tools/SPIRV-Tools should be optional. It may be useful but it is also heavy, and I also saw weird internal crashes in Emscripten+glslang interaction. Point is, *I must be able to upload GLSL shaders to OpenGL backend of Diligent "as is"*.

### My problem

- I must be able to compile shaders in runtime.
- I must be able to upload GLSL shaders to OpenGL backend of Diligent "as is".
- How on Earth do I create Diligent PipelineState then?..

On one hand, I need to provide vertex layout and immutable samplers in order to fill `GraphicsPipelineStateCreateInfo` and create `IPipelineState`.

On the other hand, I have no idea what *are* the samplers and the vertex layout until I link the shader program... which happens during `IPipelineState` creation. (I am ignoring "separable programs" since it is an optional extension.)

Therefore, it is basically _impossible_ to use OpenGL backend of Diligent with arbitrary GLSL shader. I have to somehow deduce vertex input slots and samplers on my own from raw shader code.

### My solution

I added OpenGL-specific hook that let's me modify `GraphicsPipelineStateCreateInfo` or `ComputePipelineStateCreateInfo` *after* I initiated pipeline state creation and Diligent has linked the program. This way, I can at least write OpenGL-specific code and obtain all the information I need.

This... is ugly as hell. Thanks to OpenGL being retarded. However, it enables the use case that was outright _impossible_ before. And this hack is limited to OpenGL backend, so it doesn't contaminate the rest of the library.

It would be great to have better solution, but I couldn't come up with one.

PS: I am creating this PR to discuss this issue and to share my solution. I don't really care if this tweak goes to the upstream or not.

### DIFF
--- a/Graphics/GraphicsEngine/interface/PipelineState.h
+++ b/Graphics/GraphicsEngine/interface/PipelineState.h
@@ -717,6 +717,9 @@ struct PipelineStateCreateInfo
 typedef struct PipelineStateCreateInfo PipelineStateCreateInfo;
 
 
+/// Callback that is invoked after OpenGL program object is linked.
+typedef void(DILIGENT_CALL_TYPE* GLProgramLinkedCallbackType)(Uint32* ProgramObjects, Uint32 NumProgramObjects, void* pUserData);
+
 /// Graphics pipeline state initialization information.
 struct GraphicsPipelineStateCreateInfo DILIGENT_DERIVE(PipelineStateCreateInfo)
 
@@ -743,6 +746,13 @@ struct GraphicsPipelineStateCreateInfo DILIGENT_DERIVE(PipelineStateCreateInfo)
 
     /// Mesh shader to be used with the pipeline.
     IShader* pMS DEFAULT_INITIALIZER(nullptr);
+
+    /// OpenGL only: Callback that is executed when program(s) are linked.
+    /// It is allowed to modify CreateInfo in this callback.
+    GLProgramLinkedCallbackType GLProgramLinkedCallback DEFAULT_INITIALIZER(nullptr);
+
+    /// OpenGL only: user data for GLProgramLinkedCallback.
+    void* GLProgramLinkedCallbackUserData DEFAULT_INITIALIZER(nullptr);
 
 #if DILIGENT_CPP_INTERFACE
     bool operator==(const GraphicsPipelineStateCreateInfo& Rhs) const noexcept
@@ -775,6 +785,13 @@ struct ComputePipelineStateCreateInfo DILIGENT_DERIVE(PipelineStateCreateInfo)
 
     /// Compute shader to be used with the pipeline
     IShader* pCS DEFAULT_INITIALIZER(nullptr);
+
+    /// OpenGL only: Callback that is executed when program(s) are linked.
+    /// It is allowed to modify CreateInfo in this callback.
+    GLProgramLinkedCallbackType GLProgramLinkedCallback DEFAULT_INITIALIZER(nullptr);
+
+    /// OpenGL only: user data for GLProgramLinkedCallback.
+    void* GLProgramLinkedCallbackUserData DEFAULT_INITIALIZER(nullptr);
 
 #if DILIGENT_CPP_INTERFACE
     ComputePipelineStateCreateInfo() noexcept
--- a/Graphics/GraphicsEngine/src/PSOSerializer.cpp
+++ b/Graphics/GraphicsEngine/src/PSOSerializer.cpp
@@ -201,7 +201,7 @@ bool PSOSerializer<Mode>::SerializeCreateInfo(
 
     // Skip NodeMask
 
-    ASSERT_SIZEOF64(GraphicsPipelineStateCreateInfo, 344, "Did you add a new member to GraphicsPipelineStateCreateInfo? Please add serialization here.");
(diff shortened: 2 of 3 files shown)


TheMostDiligent:
TBH, I don't understand the problem.

> I must be able to compile shaders in runtime.

I see no issues here - just go for it

> I must be able to upload GLSL shaders to OpenGL backend of Diligent "as is".

Use [`SHADER_SOURCE_LANGUAGE_GLSL_VERBATIM`](https://github.com/DiligentGraphics/DiligentCore/blob/e7ca89804517df43e6154b4e16c018d212cdadee/Graphics/GraphicsEngine/interface/Shader.h#L67)

> How on Earth do I create Diligent PipelineState then?..

You compile the shader and create the pipeline as usual

> For OpenGL backend, using glslang/SPIRV-Tools/SPIRV-Tools should be optional

glslang/SPIRV-Tools/SPIRV-Tools are not used in OpenGL backend, the shaders are compiled directly by the dirver. You can remove all shader compilation tools using `DILIGENT_NO_GLSLANG` and `DILIGENT_NO_HLSL` [build options](DILIGENT_NO_HLSL)

> On one hand, I need to provide vertex layout and immutable samplers in order to fill GraphicsPipelineStateCreateInfo and create IPipelineState.

This is after you compiled your shader. You also don't have to use immutable samplers and can set [samplers in texture views](https://github.com/DiligentGraphics/DiligentCore/blob/e7ca89804517df43e6154b4e16c018d212cdadee/doc/TextureSamplers.md).
 
> On the other hand, I have no idea what are the samplers and the vertex layout until I link the shader program... which happens during IPipelineState creation. (I am ignoring "separable programs" since it is an optional extension.)

I don't understand this at all. You don't know what shader you are compiling and which resources it uses? Do you need to get shader reflection information to automatically create the vertex layout?

I am having hard times to understand the problem here, but I may guess that what you need is the ability to create Diligent shader from exiting OpenGL program handle. This way you can create the programs the way you want, Diligent will only attach to the existing handle and load all resources.

eugeneko:
> This is after you compiled your shader. You also don't have to use immutable samplers

On the other hand, I **must** use immutable samplers on DX12 backend because it has only 2048 mutable samplers which will be spent instantly. If I use only "dynamic" samplers for OpenGL and only "immutable" samplers for DX12, I will have different high-level Diligent code for different backends... And I try to avoid that.

I initially implemented this tweak to deal with input layout, immutable samplers are just a useful side-effect tbh.

> I don't understand this at all. You don't know what shader you are compiling and which resources it uses?

I may know something about the shader, but the `LayoutElement.InputIndex` is undefined until program object is linked, so it is _impossible_ to fill `LayoutElement`s beforehand if the application knows only names of the vertex inputs.

> I am having hard times trying to understand the problem here, but I may guess that what you need is the ability to create Diligent shader from exiting OpenGL program handle.

That will also work and I even tried to do that, but it required more changes in Diligent than I was comfortable with.
I do confirm that being able to link the program myself and use reflection from it to create pipeline state will work for me.
However, pipeline state *must* reuse the same program object I used for reflection, it shall not create another one.

TheMostDiligent:
So per our discussion in Discord, here is how your problem can be tackled now:

* Create a Diligent shader object dynamically at run time
* Before you create your pipeline state, collect all shaders the pipeline will use
  * Use [`IShaderGL::GetGLShaderHandle`](https://github.com/DiligentGraphics/DiligentCore/blob/194bbd7045d0984c1cfdd1a8bf955fd8e932e317/Graphics/GraphicsEngineOpenGL/interface/ShaderGL.h#L52) for each shader object to obtain native OpenGL handles
  * Manually link all shaders into a program
  * Use OpenGL shader reflection to query the vertex shader inputs with names and locations
  * Define `InputLayout` using the reflection information
* Use the `InputLayout` to create the Pipeline State
* Use [`IPipelineStateGL::GetGLProgramHandle(SHADER_TYPE_VERTEX)`](https://github.com/DiligentGraphics/DiligentCore/blob/194bbd7045d0984c1cfdd1a8bf955fd8e932e317/Graphics/GraphicsEngineOpenGL/interface/PipelineStateGL.h#L63C27-L63C45) to get the handle of the program
  * Use `glBindAttribLocation` to remap inputs in the program to match the `InputLayout` you defined (to make sure that inputs are consistent with layout)

eugeneko:
@TheMostDiligent I can confirm that your solution works for me, and it was quite easy to migrate the code.

PR #447, opened 2024-01-05
Sampler*Impl: Validate `MaxAnisotropy` value

Arcnor:
Vulkan has validation for the `MaxAnisotropy` value, but other renderers don't, although they have similar requirements.

I've tested the code for D3D11 and D3D12, but haven't been able to test with OpenGL.


Graphics/GraphicsEngineOpenGL/src/SamplerGLImpl.cpp
```diff
@@ -102,7 +102,15 @@ SamplerGLImpl::SamplerGLImpl(IReferenceCounters* pRefCounters, class RenderDevic
     }
 
     if (SamPrpos.AnisotropicFilteringSupported) // Can be unsupported
+    {
+#if defined(DILIGENT_DEBUG) || defined(DILIGENT_DEVELOPMENT)
+        float maxLevel;
+        glGetFloatv(GL_MAX_TEXTURE_MAX_ANISOTROPY_EXT, &maxLevel);
```
TheMostDiligent, line 108: This will produce error on all platforms that don't support the `GL_MAX_TEXTURE_MAX_ANISOTROPY_EXT` extension

TheMostDiligent (commented):
- I suggest adding a single check for max anisotropy value [here](https://github.com/DiligentGraphics/DiligentCore/blob/master/Graphics/GraphicsEngine/src/SamplerBase.cpp#L42)
- Why are [HLSLDefinitions_inc.fxh](https://github.com/DiligentGraphics/DiligentCore/pull/447/files#diff-f7a39c0f8a95f051d0d031671a0b405bba48e81fbd70f7abef7d007932d38f46) and [GLSLDefinitions_inc.h](https://github.com/DiligentGraphics/DiligentCore/pull/447/files#diff-2b113fa3cd96a138d37040304d71b32562ebaa6f2826303d59980a6df8935402) in the PR?

Arcnor:
> * I suggest adding a single check for max anisotropy value [here](https://github.com/DiligentGraphics/DiligentCore/blob/master/Graphics/GraphicsEngine/src/SamplerBase.cpp#L42)

Hmm, but the limits are different per backend, right?

> * Why are [HLSLDefinitions_inc.fxh](https://github.com/DiligentGraphics/DiligentCore/pull/447/files#diff-f7a39c0f8a95f051d0d031671a0b405bba48e81fbd70f7abef7d007932d38f46) and [GLSLDefinitions_inc.h](https://github.com/DiligentGraphics/DiligentCore/pull/447/files#diff-2b113fa3cd96a138d37040304d71b32562ebaa6f2826303d59980a6df8935402) in the PR?

Sorry, I messed up while amending the PR, will remove soon.

TheMostDiligent:
> Hmm, but the limits are different per backend, right?

It is hard-coded to 16 on Direct3D. I don't think any hardware support anisotropy level higher than this.

To properly check max anisotropy on OpenGL, you need to first check if the extension is supported (it is different in OpenGL and GLES) and only then query the maximum value.

Arcnor:
> > Hmm, but the limits are different per backend, right?
> 
> It is hard-coded to 16 on Direct3D. I don't think any hardware supports anisotropy level higher than this.
> 
> To properly check max anisotropy on OpenGL, you need to first check if the extension is supported (it is different in OpenGL and GLES) and only then query the maximum value.

Ok, will make this change now as a generic check instead, although I have one last question: `IsAnisotropicFilter` is defined as:
https://github.com/DiligentGraphics/DiligentCore/blob/a28ae843ed9977559b3ed50f11fe701e17e5dcc6/Graphics/GraphicsAccessories/interface/GraphicsAccessories.hpp#L462-L468

However, according to the docs (and confusingly enough...) at https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_sampler_desc the `MaxAnisotropy` value will only be checked for `D3D12_FILTER_ANISOTROPIC` or `D3D12_FILTER_COMPARISON_ANISOTROPIC`. Is that a bug in the function, or in the MS docs?

TheMostDiligent:
A real fix here is to replace the [`AnisotropicFilteringSupported`](https://github.com/DiligentGraphics/DiligentCore/blob/master/Graphics/GraphicsEngine/interface/GraphicsTypes.h#L2075) member of the `SamplerProperties` struct with `MaxAnistropy` and check the value against it.

TheMostDiligent:
Implemented in e94b36978ccf8dd6e48c759318ef1b887496a7c5

TheMostDiligent:
If your concern is that the page you mentioned don't refer to min/max anisotropic filters, this is most certainly because Microsoft forgot to update the page.

PR #557, opened 2024-07-11
Optimize VAOCache.

eugeneko:
std::unordered_multimap emplace can be O(N) when there are a lot of repeated keys. It causes visible lag in Web when number of VAOs exceed several hundred.

### DIFF
--- a/Graphics/GraphicsEngineOpenGL/include/VAOCache.hpp
+++ b/Graphics/GraphicsEngineOpenGL/include/VAOCache.hpp
@@ -85,8 +85,8 @@ private:
         // PSO uniquely defines the layout (attrib pointers, divisors, etc.),
         // so we do not need to add individual layout elements to the key.
         // The key needs to contain all bound buffers.
-        const UniqueIdentifier PsoUId;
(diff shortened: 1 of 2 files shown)


Graphics/GraphicsEngineOpenGL/include/VAOCache.hpp
```diff
@@ -85,8 +85,8 @@ class VAOCache
         // PSO uniquely defines the layout (attrib pointers, divisors, etc.),
         // so we do not need to add individual layout elements to the key.
         // The key needs to contain all bound buffers.
-        const UniqueIdentifier PsoUId;
-        const UniqueIdentifier IndexBufferUId;
+        UniqueIdentifier PsoUId;
+        UniqueIdentifier IndexBufferUId;
```
TheMostDiligent, line 89: Can you add default initialization for these members to make sure there is no garabage?

eugeneko: What about `Streams`? There is a comment "Do not zero-out", I assumed the lack of default initialization is intentional.

PR #570, opened 2024-07-18
Egl non owned (Android)

triplejam:
Attach to an already existing EGL context.

### DIFF
--- a/Graphics/GraphicsEngineOpenGL/src/GLContextAndroid.cpp
+++ b/Graphics/GraphicsEngineOpenGL/src/GLContextAndroid.cpp
@@ -43,12 +43,21 @@
 namespace Diligent
 {
 
-bool GLContext::InitEGLSurface()
+bool GLContext::InitEGLDisplay()
 {
     display_ = eglGetDisplay(EGL_DEFAULT_DISPLAY);
     if (display_ == EGL_NO_DISPLAY)
     {
         LOG_ERROR_AND_THROW("No EGL display found");
+    }
+    return true;
+}
+
+bool GLContext::InitEGLSurface()
+{
+    if (!window_)
+    {
+        LOG_ERROR_AND_THROW("Native window pointer must be passed to EngineCreateInfo for Diligent to make EGL Surface");
     }
 
     auto success = eglInitialize(display_, &egl_major_version_, &egl_minor_version_);
@@ -123,8 +132,7 @@ bool GLContext::InitEGLSurface()
         LOG_ERROR_AND_THROW("Failed to create EGLSurface");
     }
 
-    eglQuerySurface(display_, surface_, EGL_WIDTH, &screen_width_);
-    eglQuerySurface(display_, surface_, EGL_HEIGHT, &screen_height_);
+    UpdateScreenSize();
 
     /* EGL_NATIVE_VISUAL_ID is an attribute of the EGLConfig that is
      * guaranteed to be accepted by ANativeWindow_setBuffersGeometry().
@@ -194,6 +202,15 @@ bool GLContext::InitEGLContext()
     return true;
 }
 
+void GLContext::AttachToCurrentEGLSurface()
+{
+    if (eglGetCurrentSurface(EGL_DRAW) == EGL_NO_SURFACE)
+    {
+        LOG_ERROR_AND_THROW("Failed to attach to EGLSurface: no active surface");
+    }
+    UpdateScreenSize();
+}
+
 void GLContext::AttachToCurrentEGLContext()
 {
     if (eglGetCurrentContext() == EGL_NO_CONTEXT)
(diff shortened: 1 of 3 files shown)


triplejam:
Quick SDL2 polling events pseudocode:

```cpp
while (SDL_PollEvent(&event))
{
    switch(event.type)
    {
        case SDL_WINDOWEVENT:
        {
             switch(event.window.event) {
             case SDL_WINDOWEVENT_FOCUS_LOST: {
                        (GLESDevice*)device->OnSuspend();
                        break;
              }
              case SDL_WINDOWEVENT_FOCUS_GAINED: {
                        (IGLESDevice*)device->OnResume();
                        break;
                    }
                    case SDL_WINDOWEVENT_RESIZED:
                    {
                        (IGLESDevice*)device->OnSuspend();
                        (IGLESDevice*)device->OnResume();
                        break;
                    }
                }
                break;
            }
    }
}```

File: Platforms/Apple/src/AppleDebug.mm [hunk 1]
```diff
@@ -58,7 +58,7 @@
             }
             else
             {
-                return false;
+                return (signed char)false;
```
TheMostDiligent, line 61: I think this change made it to PR by accident.

Graphics/GraphicsEngineOpenGL/src/GLContextAndroid.cpp
```diff
@@ -142,6 +152,19 @@ bool GLContext::InitEGLSurface()
...
+    {
+        glGetIntegerv(GL_MAJOR_VERSION, &major_version_);
+        glGetIntegerv(GL_MINOR_VERSION, &minor_version_);
+        context_ = eglGetCurrentContext();
+        if (context_ == EGL_NO_CONTEXT)
+        {
+            LOG_ERROR_AND_THROW("Failed to attach to existing EGL context");
+        }
+        context_valid_ = true;
+        return true;
```
TheMostDiligent, line 165: There is `AttachToCurrentEGLContext()` function that does the same

TheMostDiligent:
There is [HelloAR](https://github.com/DiligentGraphics/DiligentSamples/tree/master/Android/HelloAR) tutorial that uses context initialized by the application - maybe this is what you need?

Platforms/Apple/src/AppleDebug.mm
triplejam: It's unrelated, but I build on OSX too and this was needed for the latest clang.

triplejam:
> There is [HelloAR](https://github.com/DiligentGraphics/DiligentSamples/tree/master/Android/HelloAR) tutorial that uses context initialized by the application - maybe this is what you need?

I saw that, but it doesn't create the full swapchain. I am not aware of anyway to do that. My intent is for behaviour similar to the glfw example where it just attaches to the context created by glfw.

Graphics/GraphicsEngineOpenGL/src/GLContextAndroid.cpp
triplejam: It doesn't or its mislabeled. It assumes that you are checking for the presence of the one created by diliegent engine. It needs to be reworked in my opinion if that is the intention.

triplejam:
The one thing I am not sure about is if getting the swap intervals is technically correct.

Graphics/GraphicsEngineOpenGL/src/GLContextAndroid.cpp
```diff
@@ -233,14 +253,16 @@ bool GLContext::Init(ANativeWindow* window)
...
     //
     window_ = window;
-    if (window != nullptr)
+    InitEGLDisplay();
+    if (eglGetCurrentContext() != EGL_NO_CONTEXT)
     {
-        InitEGLSurface();
-        InitEGLContext();
+        AttachToCurrentEGLSurface();
+        AttachToCurrentEGLContext();
```
TheMostDiligent, line 260: Does it make sense to check the surface and context independently?
E.g., maybe there is context but no surface?

triplejam: I thought about that. Do you think AttachToCurrentEGL surface should be allowed to fail? And then try to create it?

Platforms/Apple/src/AppleDebug.mm
triplejam: Will remove this if you want. Otherwise I don't see the point of a separate pull request.

triplejam:
Just ran an official example and noticed a bug. So not done yet.

triplejam:
I think you will need a test for the idiopathic cases where only a context or a surface is provided.

Graphics/GraphicsEngineOpenGL/src/GLContextAndroid.cpp
```diff
@@ -342,10 +389,17 @@ void GLContext::Terminate()
 
 void GLContext::UpdateScreenSize()
 {
+    EGLSurface CurrSurface = (surface_ == EGL_NO_SURFACE) ? eglGetCurrentSurface(EGL_DRAW) : surface_;
+
+    if (CurrSurface == EGL_NO_SURFACE)
+    {
+        LOG_ERROR_AND_THROW("Failed to attach to EGLSurface: no active surface");
```
TheMostDiligent, line 396: We through exception only from constructors during initialization. Since this function can be called at run time, let's change this to error message.

TheMostDiligent (approved):
Overall, this looks good to me.

TheMostDiligent:
@triplejam Can you please as part of this commit create a document about the integration with SDL2 [here](https://github.com/DiligentGraphics/DiligentCore/tree/master/doc)?

Graphics/GraphicsEngineOpenGL/src/GLContextAndroid.cpp
triplejam: Okay. I changed it to log the error and return early.

triplejam:
> @triplejam Can you please as part of this PR create a document about the integration with SDL2 [here](https://github.com/DiligentGraphics/DiligentCore/tree/master/doc)?

Yes. It may not be completely exhaustive but I can do.

PR #573, opened 2024-07-20
WebGPU: Implemented contant buffer reflection (close #503)

### LINKED ISSUE #503: WebGPU: implement constant buffer reflection
Use [SPIRVShaderResources](https://github.com/DiligentGraphics/DiligentCore/blob/b7f9b0e7d3c636a2127c9aabc5fb13321b3df016/Graphics/ShaderTools/src/SPIRVShaderResources.cpp#L341-L360) as example
Tests: https://github.com/DiligentGraphics/DiligentCore/blob/master/Tests/DiligentCoreAPITest/src/ConstantBufferReflectionTest.cpp

MikhailGorobets:


TheMostDiligent:
* Vulkan tests fail: row-major matrices need to be properly detected
* To avoid changing the offsets in tests, replace 4x2 matrices with 2x4 matrices

PR #601, opened 2024-08-07
WebGPU: Reworked map/unmap (close #597, close #596)

### LINKED ISSUE #596: WebGPU: implement fences
Requires: #597 

There is no general way to track GPU execution in WebGPU, but we need to support one special case: copying to a staging texture/buffer and mapping it for reading.
Handling this special use case transparently for the client can be implemented as follows:

For each signaled value, a fence keeps a list of sync points. The sync point is a simple class derived from IObject that only holds one atomic bool value.

When a buffer or texture is mapped for reading, the async callback creates a sync point and adds it to the device context. Device context accumulates all pending sync points until the next flush.

When a fence is signaled, it is added to the list of pending fences in the context.

When flush happens, all sync points from device context are added to all pending fences.
The context discards sync points that are signaled.

When async callback is called, it signals the sync point.

To get current value, the fence checks all sync points for the value, if they all are signaled, the value is updated.

⚠️ Async callback must keep a strong reference to the sync point so that it can be safely fired after device and all objects are released.

When the callback is called, staging buffer/texture should copy data to a temporary buffer to then allow client call

Other changes that need to be made here:

- Remove busy waits in Buffer/Texture
```
        while (wgpuBufferGetMapState(m_wgpuBuffer.Get()) != WGPUBufferMapState_Unmapped)
            m_pDevice->DeviceTick();
```

- Rework `DeviceContext::WaitForIdle` by issuing DeviceTick and waiting for all sync points to signal.
  - `WaitForIdle` will be disabled on the Web

### LINKED ISSUE #597: WebGPU: refactor readback buffer/textures
To allow staging reaback buffer/textures behave similarly to other APIs, refactor them as follows:

- When data is copied to staging buffer/texture, add the buffer/texture to the context's list of dirty staging buffer/textures
- When command buffer is flushed, iterate through the list and call `MapAsync` for each one
  - When callback is called, copy the data to CPU memory
  - ⚠️ It is important to make sure that a buffer does not mapped multiple times. It will be necessary to create multiple internal buffers if current is mapped, but some command (e.g. Copy) needs a buffer
  - ⚠️ The callback must hold strong reference to the buffer to make sure it is alive in case the callback is called after the buffer is released
- Mapping buffer for reading will simply return the pointer to internal CPU copy

MikhailGorobets:


Graphics/GraphicsEngineWebGPU/include/SyncPointWebGPU.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Diligent
+{
+
+struct ISyncPointWebGPU : public IObject
+{
+    virtual Bool GetValue() const = 0;
+
+    virtual void SetValue(Bool Value) = 0;
+};
```
TheMostDiligent, line 43: Interface is not needed. Sync point is an internal implementation detail that will not be exposed to clients.

Graphics/GraphicsEngineWebGPU/src/BufferWebGPUImpl.cpp
```diff
@@ -350,6 +261,53 @@ void BufferWebGPUImpl::SetDynamicAllocation(DeviceContextIndex CtxId, DynamicMem
...
+{
+    auto MapAsyncCallback = [](WGPUBufferMapAsyncStatus MapStatus, void* pUserData) {
+        if (MapStatus != WGPUBufferMapAsyncStatus_Success && MapStatus != WGPUBufferMapAsyncStatus_DestroyedBeforeCallback)
+            DEV_ERROR("Failed wgpuBufferMapAsync: ", MapStatus);
+
+        if (MapStatus == WGPUBufferMapAsyncStatus_Success && pUserData != nullptr)
+        {
+            auto* pBufferInfo = static_cast<StagingBufferSyncInfo*>(pUserData);
+
+            const auto* pData = static_cast<const uint8_t*>(wgpuBufferGetConstMappedRange(pBufferInfo->wgpuBuffer, 0, WGPU_WHOLE_MAP_SIZE));
```
TheMostDiligent, line 300: ⚠️ This is the source of potential problem: the data passed to the callback must be alive and potentially outlive the buffer itself.
To ensure this:
- Make the data derived from IObject
- Manually call AddRef before passing data to the callback
- Manually call Release in the callback

Graphics/GraphicsEngineWebGPU/src/BufferWebGPUImpl.cpp
```diff
@@ -381,4 +339,60 @@ void BufferWebGPUImpl::CreateViewInternal(const BufferViewDesc& OrigViewDesc, IB
...
+    {
+        WGPUBufferDescriptor wgpuBufferDesc{};
+        wgpuBufferDesc.label            = m_Desc.Name;
+        wgpuBufferDesc.size             = AlignUp(m_Desc.Size, m_Alignment);
+        wgpuBufferDesc.usage            = WGPUBufferUsage_MapWrite | WGPUBufferUsage_CopySrc;
+        wgpuBufferDesc.mappedAtCreation = true;
+
+        WebGPUBufferWrapper wgpuBuffer{wgpuDeviceCreateBuffer(m_pDevice->GetWebGPUDevice(), &wgpuBufferDesc)};
+        if (!wgpuBuffer)
+            LOG_ERROR_AND_THROW("Failed to create WebGPU buffer ", " '", m_Desc.Name ? m_Desc.Name : "", '\'');
```
TheMostDiligent, line 354: Errors may only be thrown from constructors.
Print an error here and return.

Graphics/GraphicsEngineWebGPU/src/BufferWebGPUImpl.cpp
```diff
@@ -381,4 +339,60 @@ void BufferWebGPUImpl::CreateViewInternal(const BufferViewDesc& OrigViewDesc, IB
...
+    }
+
+    WGPUBufferDescriptor wgpuBufferDesc{};
+    wgpuBufferDesc.label = m_Desc.Name;
+    wgpuBufferDesc.size  = AlignUp(m_Desc.Size, m_Alignment);
+    wgpuBufferDesc.usage = WGPUBufferUsage_MapRead | WGPUBufferUsage_CopyDst;
+
+    WebGPUBufferWrapper wgpuBuffer{wgpuDeviceCreateBuffer(m_pDevice->GetWebGPUDevice(), &wgpuBufferDesc)};
+    if (!wgpuBuffer)
+        LOG_ERROR_AND_THROW("Failed to create WebGPU buffer ", " '", m_Desc.Name ? m_Desc.Name : "", '\'');
```
TheMostDiligent, line 384: No throws besides from constructors.

Graphics/GraphicsEngineWebGPU/src/BufferWebGPUImpl.cpp
```diff
@@ -199,88 +192,41 @@ WGPUBuffer BufferWebGPUImpl::GetWebGPUBuffer() const
     return m_pDevice->GetDynamicMemoryManager().GetWGPUBuffer();
 }
 
-void BufferWebGPUImpl::Map(MAP_TYPE MapType, MAP_FLAGS MapFlags, PVoid& pMappedData)
+Uint64 BufferWebGPUImpl::GetDynamicOffset(DeviceContextIndex CtxId, const DeviceContextWebGPUImpl* pCtx) const
```
TheMostDiligent, line 195: This function is kept in the header to help compiler inline it.
Do not mix changes that are not related to the topic of PR.

Graphics/GraphicsEngineWebGPU/include/BufferWebGPUImpl.hpp
```diff
@@ -46,6 +47,15 @@ class BufferWebGPUImpl final : public BufferBase<EngineWebGPUImplTraits>
 public:
     using TBufferBase = BufferBase<EngineWebGPUImplTraits>;
 
+    struct StagingBufferSyncInfo
```
TheMostDiligent, line 50: ⚠️ This data is passed to the callback. It needs to be derived from IObject and AddRef'ed before it is passed to the callback.

Graphics/GraphicsEngineWebGPU/include/DeviceContextWebGPUImpl.hpp
```diff
@@ -584,6 +569,38 @@ class DeviceContextWebGPUImpl final : public DeviceContextBase<EngineWebGPUImplT
...
+        enum ResourceType
+        {
+            Buffer,
+            Texture
+        };
+
+        RefCntAutoPtr<IDeviceObject>    pResource;
+        RefCntAutoPtr<ISyncPointWebGPU> pSyncPoint;
+        ResourceType                    ResourceType;
+        Uint32                          ResourceIdentifier;
```
TheMostDiligent, line 583: Always default-initialize basic types

Graphics/GraphicsEngineWebGPU/include/FenceWebGPUImpl.hpp
```diff
@@ -56,7 +57,14 @@ class FenceWebGPUImpl final : public FenceBase<EngineWebGPUImplTraits>
...
 
-    void SetCompletedValue(Uint64 Value);
+    void AppendSyncPoints(const std::vector<RefCntAutoPtr<ISyncPointWebGPU>>& SyncPoints, Uint64 EventOnComplete);
+
+    void RequestedValue(Uint64 Value);
+
+private:
+    using SyncPointEvent = std::pair<Uint64, std::vector<RefCntAutoPtr<ISyncPointWebGPU>>>;
+    std::atomic<Uint64>        m_RequestedFenceValue{0};
+    std::deque<SyncPointEvent> m_SyncPoints;
```
TheMostDiligent, line 67: There should be a list of sync points for each signalled value.
Something like `std::map<Uint64, std::vector<SyncPoint>>`
Map will keep them sorted, so it will be easy to iterate.

TheMostDiligent (changes requested):

Graphics/GraphicsEngineWebGPU/src/BufferWebGPUImpl.cpp
MikhailGorobets: There's no problem with the lifetime here, because when the buffer is destroyed, a WGPUBufferMapAsyncStatus_DestroyedBeforeCallback event will be triggered.

Graphics/GraphicsEngineWebGPU/include/BufferWebGPUImpl.hpp
MikhailGorobets: No need to complicate the code; the object lifetimes are well-defined. I doubt anyone would call 16 flushes inside a frame

Graphics/GraphicsEngineWebGPU/include/FenceWebGPUImpl.hpp
MikhailGorobets: Why use a map if the deque is already sorted?

Graphics/GraphicsEngineWebGPU/src/BufferWebGPUImpl.cpp
MikhailGorobets: UPD:
I discovered a situation where `WGPUBufferMapAsyncStatus_DestroyedBeforeCallback` won't help us. When we use a `WGPUBuffer` in a `WGPUCommandList`, it implicitly increases the reference count of the object, which could lead to a situation where we receive `WGPUBufferMapAsyncStatus_Success`, but the object has already been deleted on our side. The method you suggested won't work because we also need to track the lifetime of the object itself, as `pMappedData`might be invalid. Therefore, I decided to simply store a reference to the object. This way, we can just increase the reference count when the asynchronous operation is called and decrease it when the operation is completed.

TheMostDiligent: Here is what I suggest:
Before the callback:
```
pData->AddRef();
Callback() = [](){
  // Do work
  pData->Release();
};
Callback(... pData)
```
`pData` **must** contain all data that may be accessed by the callback. This is guaranteed to work,

PR #604, opened 2024-08-09
WebGPU: Added support for formats without storage flag in mip generat…

### LINKED ISSUE #590: WebGPU: support mip generations for formats that do not allow storage access
Formats like 16-bit floats do not allow storage access and as a result, mip generation fails.

MikhailGorobets:
…or (close #590)

### DIFF
--- a/Graphics/GraphicsEngineWebGPU/include/TextureViewWebGPUImpl.hpp
+++ b/Graphics/GraphicsEngineWebGPU/include/TextureViewWebGPUImpl.hpp
@@ -61,6 +61,10 @@ public:
 
     WGPUTextureView GetMipLevelSRV(Uint32 Mip);
 
+    WGPUTextureView GetMipLevelRTV(Uint32 Slice, Uint32 Mip);
+
+    WGPUTextureView GetMipLevelSRV(Uint32 Slice, Uint32 Mip);
+
 private:
     WebGPUTextureViewWrapper              m_wgpuTextureView;
     std::vector<WebGPUTextureViewWrapper> m_wgpuTextureMipSRVs;
--- a/Graphics/GraphicsEngineWebGPU/src/TextureWebGPUImpl.cpp
+++ b/Graphics/GraphicsEngineWebGPU/src/TextureWebGPUImpl.cpp
@@ -30,6 +30,7 @@
 #include "RenderDeviceWebGPUImpl.hpp"
 #include "DeviceContextWebGPUImpl.hpp"
 #include "WebGPUTypeConversions.hpp"
+#include "GraphicsAccessories.hpp"
 
 namespace Diligent
 {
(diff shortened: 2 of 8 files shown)


MikhailGorobets:
Can't reproduce issue on WSL

Graphics/GraphicsEngineWebGPU/include/TextureViewWebGPUImpl.hpp
```diff
@@ -61,6 +61,10 @@ class TextureViewWebGPUImpl final : public TextureViewBase<EngineWebGPUImplTrait
 
     WGPUTextureView GetMipLevelSRV(Uint32 Mip);
 
+    WGPUTextureView GetMipLevelRTV(Uint32 Slice, Uint32 Mip);
+
+    WGPUTextureView GetMipLevelSRV(Uint32 Slice, Uint32 Mip);
+
 private:
     WebGPUTextureViewWrapper              m_wgpuTextureView;
     std::vector<WebGPUTextureViewWrapper> m_wgpuTextureMipSRVs;
```
TheMostDiligent, line 70: These views should be in the Texture object

Graphics/GraphicsEngineWebGPU/src/TextureWebGPUImpl.cpp
```diff
@@ -66,10 +69,22 @@ WGPUTextureDescriptor TextureDescToWGPUTextureDescriptor(const TextureDesc&
...
+    {
+        if (FmtInfo.BindFlags & BIND_UNORDERED_ACCESS)
+            wgpuTextureDesc.usage |= WGPUTextureUsage_StorageBinding;
+        else if (FmtInfo.BindFlags & BIND_RENDER_TARGET)
+            wgpuTextureDesc.usage |= WGPUTextureUsage_RenderAttachment;
+        else
+            LOG_ERROR_AND_THROW("Automatic mipmap generation isn't supported for ", GetTextureFormatAttribs(Desc.Format).Name, " as the format can't be used as render target or storage texture");
+
+        if (!FmtInfo.Filterable)
+            LOG_ERROR_AND_THROW("Automatic mipmap generation isn't supported for ", GetTextureFormatAttribs(Desc.Format).Name, " as the format doesn't support linear filtering");
```
TheMostDiligent, line 86: You can manually filter 4 values in the shader, this way this limitation can be lifted

TheMostDiligent (changes requested):

Graphics/GraphicsEngineWebGPU/include/TextureViewWebGPUImpl.hpp
MikhailGorobets: I specifically asked about this when we first started working on it, and we decided that the logic would be like in D3D12

Graphics/GraphicsEngineWebGPU/src/TextureWebGPUImpl.cpp
MikhailGorobets: We have the same logic in the Vulkan backend
https://github.com/DiligentGraphics/DiligentCore/blob/a56f176f6afa1adefa1acc9bd7c0afd8891ae692/Graphics/GraphicsEngineVulkan/src/TextureVkImpl.cpp#L118
I think this is not critical because mip generation will only fail for FP32 formats if the device doesn't support this extension `float32-filterable`

TheMostDiligent (approved):

PR #608, opened 2024-08-13
Fix return type.

triplejam:
I can't test this at the moment but a cast was need building on OSX. I think using the BOOL macro / typedef is the correct way to do it.


Platforms/Apple/src/AppleDebug.mm
```diff
@@ -58,7 +58,7 @@
             }
             else
             {
-                return false;
+                return (BOOL)false;
```
TheMostDiligent, line 61: A better way to fix this would be to make `bool` the return type of lambda

TheMostDiligent (changes requested):

PR #629, opened 2024-10-14
IDeviceContextGL: OpenGL interop improvements.

WangHoi:
Add `PurgeCurrentContextCaches()` method;
Remove the `PurgeCaches` parameter of  `UpdateCurrentGLContext()` method.

### DIFF
--- a/Graphics/GraphicsEngineOpenGL/src/RenderDeviceGLImpl.cpp
+++ b/Graphics/GraphicsEngineOpenGL/src/RenderDeviceGLImpl.cpp
@@ -1687,11 +1687,23 @@ void RenderDeviceGLImpl::PurgeContextCaches(GLContext::NativeGLContextType Conte
 {
     {
         Threading::SpinLockGuard FBOCacheGuard{m_FBOCacheLock};
-        m_FBOCache.erase(Context);
+
+        auto it = m_FBOCache.find(Context);
+        if (it != m_FBOCache.end())
+        {
+            it->second.Clear();
+            m_FBOCache.erase(it);
+        }
     }
     {
         Threading::SpinLockGuard VAOCacheGuard{m_VAOCacheLock};
-        m_VAOCache.erase(Context);
+
+        auto it = m_VAOCache.find(Context);
+        if (it != m_VAOCache.end())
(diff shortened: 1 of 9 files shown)


Graphics/GraphicsEngineOpenGL/src/RenderDeviceGLImpl.cpp
```diff
@@ -1687,11 +1687,23 @@ void RenderDeviceGLImpl::PurgeContextCaches(GLContext::NativeGLContextType Conte
 {
     {
         Threading::SpinLockGuard FBOCacheGuard{m_FBOCacheLock};
-        m_FBOCache.erase(Context);
+
+        auto it = m_FBOCache.find(Context);
+        if (it != m_FBOCache.end())
+        {
+            it->second.Clear();
```
TheMostDiligent, line 1694: Why do you need to clear the cache before erasing it?
When the cache is destroyed all objects should be released anyway.

Graphics/GraphicsEngineOpenGL/src/RenderDeviceGLImpl.cpp
```diff
@@ -1687,11 +1687,23 @@ void RenderDeviceGLImpl::PurgeContextCaches(GLContext::NativeGLContextType Conte
...
+        }
     }
     {
         Threading::SpinLockGuard VAOCacheGuard{m_VAOCacheLock};
-        m_VAOCache.erase(Context);
+
+        auto it = m_VAOCache.find(Context);
+        if (it != m_VAOCache.end())
+        {
+            it->second.Clear();
```
TheMostDiligent, line 1704: Same here - when VAO cache is destroyed in the next line, all objects it holds should be released.

WangHoi: If users `PurgeCaches`, force clearing  to avoid assertion errors in destructors of VAOCache and FBOCache.

So if VAO cache is destroyed automatically, and some internal VAO/FBO state cache bugs occurs, the assertion in its destructor can still check for possible errors.

PR #630, opened 2024-10-20
Restore Android SDK 21 support.

eugeneko:
This is standard approach for manual pointer alignment, used in e.g. EASTL:

https://github.com/electronicarts/EASTL/blob/master/include/EASTL/allocator.h#L252-L296

This fallback allows us to avoid pushing minimal required Android SDK version for sake of single optional feature.

I did not test it on Android, but I tested these functions by calling them unconditionally in `AllocateAligned` and `FreeAligned`.


BuildTools/Android/tests/build.gradle
```diff
@@ -9,7 +9,7 @@ android {
     defaultConfig {
         applicationId 		'DE.Android.Tests'
         targetSdkVersion 	32
-        minSdkVersion 		28
+        minSdkVersion 		21
```
TheMostDiligent, line 12: Let's keep 28 as the default version.

Common/src/DefaultRawMemoryAllocator.cpp
```diff
@@ -39,6 +39,37 @@
 namespace Diligent
 {
 
+namespace
+{
+
+#if PLATFORM_ANDROID && __ANDROID_API__ < 28
+void* AllocateAlignedFallback(size_t Size, size_t Alignment)
```
TheMostDiligent, line 46: This looks good to me.
I made some refactoring to have only one `#if` condition (right now `#if PLATFORM_ANDROID && __ANDROID_API__ < 28` is repeated three times)
Can you rebase on the latest and update the branch?

TheMostDiligent (changes requested):

PR #631, opened 2024-10-23
Support ETC2 texture format.

eugeneko:
Note: I tested both native GLES3.0 and WebGL2, but I cannot test mobile Vulkan/WebGPU right now.


eugeneko:
I see that Linux tests fail but I am not sure how to fix that.

We need to somehow check whether ETC2 is supported on desktop OpenGL but I cannot find appropriate extension.

PR #636, opened 2024-11-23
Add D3D_FEATURE_LEVEL_12_2

2A5F:


TheMostDiligent:
Hello!

This change currently [fails to build](https://ci.appveyor.com/project/DiligentGraphics/diligentcore/builds/51043870/job/9te319e4b2rijoyh#L692) with the older Windows SDK version:
```
C:\projects\DiligentCore\Graphics\GraphicsEngineD3DBase\include\EngineFactoryD3DBase.hpp(298,29): error C2065: 'D3D_FEATURE_LEVEL_12_2': undeclared identifier (compiling source file C:\projects\DiligentCore\Graphics\GraphicsEngineD3D12\src\EngineFactoryD3D12.cpp) [C:\projects\DiligentCore\build\Graphics\GraphicsEngineD3D12\Diligent-GraphicsEngineD3D12-static.vcxproj]
```
It is necessary to find out in which SDK version these enum value was first defined and [add a fallback value for older SDK versions](https://github.com/DiligentGraphics/DiligentCore/blob/master/Graphics/GraphicsEngineD3D12/include/pch.h#L47-L62).

2A5F:
Should use [directx heades](https://github.com/Microsoft/DirectX-Headers) instead of windows sdk to avoid this situation

TheMostDiligent:
> Should use [directx heades](https://github.com/Microsoft/DirectX-Headers) instead of windows sdk to avoid this situation

This will require quite a bit of work, which is not justified by this specific change.
For this PR to work, adding a fallback value is much simpler.

Graphics/GraphicsEngineD3D12/include/pch.h
```diff
@@ -58,6 +58,10 @@ enum D3D12_SHADING_RATE_COMBINER
 {
 };
 
+#ifndef NTDDI_WIN10_CO // First defined in Win SDK 10.0.22000.0
+#    define D3D_FEATURE_LEVEL_12_2 0xc200
```
TheMostDiligent, line 62: Please define it as `constexpr` constant of type `D3D_FEATURE_LEVEL` similar to `D3D12_RESOURCE_STATE_SHADING_RATE_SOURCE` below.

Also, please move this above the 10.0.19041.0 SDK to keep the defines ordered by versions.

TheMostDiligent (changes requested):

TheMostDiligent (changes requested):
It looks like required changes are missing [here](https://github.com/DiligentGraphics/DiligentCore/blob/26b091e4f6dfe573b12eadebcaf5da3383a6aee8/Graphics/GraphicsEngineD3D11/src/ShaderD3D11Impl.cpp#L63), [here](https://github.com/DiligentGraphics/DiligentCore/blob/26b091e4f6dfe573b12eadebcaf5da3383a6aee8/Graphics/GraphicsEngineD3D12/src/RenderDeviceD3D12Impl.cpp#L72), and [here](https://github.com/DiligentGraphics/DiligentCore/blob/26b091e4f6dfe573b12eadebcaf5da3383a6aee8/Graphics/GraphicsEngineD3D12/src/RenderDeviceD3D12Impl.cpp#L180)

Graphics/GraphicsEngineD3D12/include/pch.h
```diff
@@ -44,6 +44,10 @@
 
 #include "WinHPostface.h"
 
+#ifndef NTDDI_WIN10_FE // First defined in Win SDK 10.0.20348.0
+constexpr D3D_FEATURE_LEVEL D3D_FEATURE_LEVEL_12_2 = static_cast<D3D_FEATURE_LEVEL>(0xc200);
+#endif
```
TheMostDiligent, line 49: I have `10.0.22621.0`, and `D3D_FEATURE_LEVEL_12_2` is defined

TheMostDiligent: It is not defined in `10.0.19041.0`, so it looks like `10.0.22621.0` is where it was first defined

2A5F: It is defined in 10.0.20348.0

2A5F: 10.0.20348.0

```c
...
#define NTDDI_WIN10_RS5                     0x0A000006
#define NTDDI_WIN10_19H1                    0x0A000007
#define NTDDI_WIN10_VB                      0x0A000008
#define NTDDI_WIN10_MN                      0x0A000009
#define NTDDI_WIN10_FE                      0x0A00000A

#define WDK_NTDDI_VERSION                   NTDDI_WIN10_FE
```

---

10.0.22000.0

```c
...
#define NTDDI_WIN10_19H1                    0x0A000007
#define NTDDI_WIN10_VB                      0x0A000008
#define NTDDI_WIN10_MN                      0x0A000009
#define NTDDI_WIN10_FE                      0x0A00000A
#define NTDDI_WIN10_CO                      0x0A00000B

#define WDK_NTDDI_VERSION                   NTDDI_WIN10_CO
```

---

10.0.22621.0

```c
...
#define NTDDI_WIN10_FE                      0x0A00000A
#define NTDDI_WIN10_CO                      0x0A00000B
#define NTDDI_WIN10_NI                      0x0A00000C

#define WDK_NTDDI_VERSION                   NTDDI_WIN10_NI
```

TheMostDiligent: I am not saying `D3D_FEATURE_LEVEL_12_2` is not defined in 10.0.20348.0.
I am sating it _is_ defined in 10.0.22621.0, so this will fail to build with this change as the `D3D_FEATURE_LEVEL_12_2` will be double-defined.

TheMostDiligent: So the comment is then incorrect?

2A5F: Yes

Graphics/GraphicsEngineD3D11/src/ShaderD3D11Impl.cpp
```diff
@@ -54,12 +54,13 @@ static const ShaderVersion HLSLValidateShaderVersion(const ShaderVersion& Versio
 
 static const ShaderVersion GetD3D11ShaderModel(D3D_FEATURE_LEVEL d3dDeviceFeatureLevel, const ShaderVersion& HLSLVersion)
 {
-    switch (d3dDeviceFeatureLevel)
+    switch (static_cast<Uint32>(d3dDeviceFeatureLevel))
```
TheMostDiligent, line 57: I would suggest disabling this specific warning on MSVC in case 
`#ifndef NTDDI_WIN10_FE`

TheMostDiligent (changes requested):

Graphics/GraphicsEngineD3D11/src/ShaderD3D11Impl.cpp
2A5F: If so, what's the point of fall back?

Graphics/GraphicsEngineD3D11/src/ShaderD3D11Impl.cpp
```diff
@@ -54,12 +54,16 @@ static const ShaderVersion HLSLValidateShaderVersion(const ShaderVersion& Versio
 
 static const ShaderVersion GetD3D11ShaderModel(D3D_FEATURE_LEVEL d3dDeviceFeatureLevel, const ShaderVersion& HLSLVersion)
 {
+#ifdef _MSC_VER
```
TheMostDiligent, line 57: Can you please also add `!defined(NTDDI_WIN10_FE)` to this condition so that it is disabled in newer SDKs

TheMostDiligent (changes requested):

PR #637, opened 2024-11-25
Use DirectX-Headers

2A5F:


TheMostDiligent:
Currently this change is not required - it does not solve any existing problem, but complicates the build setup.
I suggest this is done when the need arises.

PR #640, opened 2024-12-10
refactor: improve MinGW detection logic in CMake

luciddream-tsin:
Replace the simple CMAKE_GENERATOR based MinGW detection with a more robust approach:
- Check if the system is Windows
- Verify compiler is either GCC or Clang with GNU frontend
- Support both MinGW-GCC and MinGW-Clang toolchains
- Fix potential false positive when using other CMake generators

This change ensures more accurate MinGW toolchain detection regardless of the CMake generator being used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,9 +176,15 @@ if(PLATFORM_WIN32 OR PLATFORM_UNIVERSAL_WINDOWS)
     endif()
 endif()
 
-if(${CMAKE_GENERATOR} MATCHES "MinGW")
-    message("Building with MinGW")
-    set(MINGW_BUILD TRUE CACHE INTERNAL "Building with MinGW")
+set(MINGW_BUILD FALSE CACHE INTERNAL "MinGW build flag")
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR
+    (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
+            CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "GNU"))
+        set(MINGW_BUILD TRUE CACHE INTERNAL "MinGW build flag")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -176,9 +176,15 @@ if(PLATFORM_WIN32 OR PLATFORM_UNIVERSAL_WINDOWS)
     endif()
 endif()
 
-if(${CMAKE_GENERATOR} MATCHES "MinGW")
-    message("Building with MinGW")
-    set(MINGW_BUILD TRUE CACHE INTERNAL "Building with MinGW")
+set(MINGW_BUILD FALSE CACHE INTERNAL "MinGW build flag")
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
```
TheMostDiligent, line 181: This is incorrect. You can cross-compile for Windows on Linux with MinGW. Try to google it.

CMakeLists.txt
```diff
@@ -176,9 +176,15 @@ if(PLATFORM_WIN32 OR PLATFORM_UNIVERSAL_WINDOWS)
...
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR
+    (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
+            CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "GNU"))
```
TheMostDiligent, line 184: Clang and gcc are certainly the most used compilers, but not the only ones.
What is the purpose of this check?

TheMostDiligent: Moreover, this check is also incorrect as you can use clang with Visual Studio build on Windows.

TheMostDiligent:
As I noted in my comments, the logic is incorrect because
* MinGW toolchain can be used on Linux to cross-compile for Windows
* Windows builds may use other compilers besides MSVC. In particular, [Visual Studio officially supports clang/LLVM](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170)

PR #642, opened 2025-01-02
String tools: make WidenString and NarrowString work with multibyte characters

WangHoi:
I found this little bug when open files which has CJK characters.


WangHoi:
Sorry, need to split ObjectBase for another PR

WangHoi:
Try to rework the PR

PR #643, opened 2025-01-02
String tools: make WidenString and NarrowString work with multibyte characters

WangHoi:


WangHoi:
@TheMostDiligent Can you have a look, some failed tests are not related to this PR.

TheMostDiligent:
They are related. The NarrowString test crashes

WangHoi:
> They are related. The NarrowString test crashes

Sorry for the crash, on which platform, which test, i'll try to fix.
I see lots of hlsl tokenizer error in CI.

WangHoi:
Not easy to fix, related to the semantics of `WidenString` and `NarrowString`.
There are three encodings: utf-8, utf-16(wide), local system.

Will WidenString converts from local system encoding to utf-16, or utf-8 to utf-16 ?

TheMostDiligent:
Here is [Linux/Release crash](https://github.com/DiligentGraphics/DiligentCore/actions/runs/12578947855/job/35058387269?pr=643#step:6:571)
Other tests fail likely because the locale is not restored in the test.

TBH, I don't like the current approach very much.
Since the goal of this PR is to fix the file open issue on Windows, a better may be to handle this inside the windows file system, assuming that the string is a UTF-8 string

WangHoi:
> Here is [Linux/Release crash](https://github.com/DiligentGraphics/DiligentCore/actions/runs/12578947855/job/35058387269?pr=643#step:6:571) Other tests fail likely because the locale is not restored in the test.
> 
> TBH, I don't like the current approach very much. Since the goal of this PR is to fix the file open issue on Windows, a better may be to handle this inside the windows file system, assuming that the string is a UTF-8 string

So, how do you think add wide string functions to `FileSystem` and `FileWrapper`, let application handle the encoding convert ?

Assume the string is a UTF-8 string may break exist code, since `WidenString` is using std::wstringstream::locale() now.

WangHoi:
Just verified the old `WidenString` logic, also broken on Linux for multibyte characters.

TheMostDiligent:
> I found no comments or documentation about the passed in string encoding, in FileWrapper like classes.

Yes, non-ASCII characters are not handled well at this point.

> So, how do you think add wide string functions to FileSystem and FileWrapper, let application handle the encoding convert ?

When you pass the "Bézier.json" string, how is it represented?

WangHoi:
> When you pass the "Bézier.json" string, how is it represented?

I always use UTF-8 in application code, using explicit utf-8 string literal when possible. 
`FileWrapper(u8"Bézier.json")`

TheMostDiligent:
I thing UTF-8 should be the default encoding when char string is passed

WangHoi:
Thanks, it seems a lot of code and doc changes.
Can you make these changes, or i provide another PR?

TheMostDiligent:
I don't really see that a lot of changes will be needed. Since ASCII is a subset of UTF-8, there should be no need for many changes.
The most important thing is to support UTF-8 file names.

> Can you make these changes, or i provide another PR?

I will think how to best do this.

WangHoi:
😄 Thank you!

PR #646, opened 2025-01-03
ObjectBase: make INTERFACE_ID fully qualified

WangHoi:
Support implement Diligent's class interface in global or other namespaces.

For example:
```cpp
/// Implementation of Diligent::IRenderStateCache
class CustomRenderStateCache final : public Diligent::ObjectBase<Diligent::IRenderStateCache>
{
public:
    using TBase = Diligent::ObjectBase<Diligent::IRenderStateCache>;

public:
    CustomRenderStateCache(Diligent::IReferenceCounters*               pRefCounters,
                         const Diligent::RenderStateCacheCreateInfo& CreateInfo);

    IMPLEMENT_QUERY_INTERFACE_IN_PLACE(Diligent::IID_RenderStateCache, TBase);
}
```

### DIFF
--- a/Common/interface/ObjectBase.hpp
+++ b/Common/interface/ObjectBase.hpp
@@ -52,11 +52,11 @@ namespace Diligent
         }                                                            \
     }
 
-#define IMPLEMENT_QUERY_INTERFACE(ClassName, InterfaceID, ParentClassName)         \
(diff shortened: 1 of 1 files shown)


TheMostDiligent:
`IMPLEMENT_QUERY_INTERFACE_IN_PLACE` is an internal macros that is not supposed to be used externally.
It is really an implementation detail.

WangHoi:
OK, close the PR.

PR #650, opened 2025-01-20
Provide example ImGui implementation for SDL

triplejam:
Not 100% complete because the backend used should depend on querying the device type. Also currently missing an assertion checking the render width and height. But works.


TheMostDiligent:
Actually, it makes sense to add the SDL implementation alongside all other implementations.

triplejam:
I will make a pull request there too then. I will rewrite this as how to use it.

TheMostDiligent:
This can now be updated to reference the SDL implementation.

Issue #124, opened 2020-03-16
Properly handle non-sRGB default framebuffer in OpenGL backend


TheMostDiligent:
Right now `GL_FRAMEBUFFER_SRGB` is set once during the engine initialization. As a result, this is always enabled for default framebuffer even when its format is not SRGB

TheMostDiligent:
OpenGL is a mess. `GL_FRAMEBUFFER_SRGB` only affects rendering to SRGB framebuffers and enables linear->sRGB conversion when writing pixels to the framebuffer. This should always be enabled to be consistent with other backends.

TheMostDiligent:
This needs to be done for default framebuffer

julianxhokaxhiu:
You may want to update the Label as well.

TheMostDiligent:
In theoretical OpenGL, `GL_FRAMEBUFFER_SRGB` should only apply to sRGB framebuffers and should not apply to non-sRGB.
In practical OpenGL, this more or less works for off-screen framebuffers, but default framebuffers are hopelessly broken.

There is [ARB_framebuffer_sRGB](https://registry.khronos.org/OpenGL/extensions/ARB/ARB_framebuffer_sRGB.txt) extension that defines `GLX_FRAMEBUFFER_SRGB_CAPABLE_ARB`/`WGL_FRAMEBUFFER_SRGB_CAPABLE_ARB` which extension authors did not even bother to describe:

> The {GLX,WGL}_FRAMEBUFFER_SRGB_CAPABLE_ARB pixel format attributes are used to select default framebuffers which are sRGB-capable in the fairly obvious way

It is anything but obvious what sRGB-capable means. Selecting non-sRGB-capable format does not work at all - the GL then fails to get the number of extensions.

Next, there is [WGL_EXT_colorspace](https://registry.khronos.org/OpenGL/extensions/EXT/WGL_EXT_colorspace.txt) extension that defines `WGL_COLORSPACE_SRGB_EXT` and `WGL_COLORSPACE_LINEAR_EXT`. It is unclear how this is supposed to interplay with `FRAMEBUFFER_SRGB_CAPABLE`, for example what does sRGB-capable framebuffer with linear color space mean?
In any case, whether pixel format with `WGL_COLORSPACE_SRGB_EXT` or `WGL_COLORSPACE_LINEAR_EXT` is selected does not matter - sRGB->Linear conversion happens all the time.
[...]

PR #675, opened 2025-05-05
OpenGL: Fix srgb framebuffer issue.

Admenri:
Fixed OpenGL swapchain TEX_FORMAT_RGBA8_UNORM setting not taking effect.

### DIFF
--- a/Graphics/GraphicsEngineOpenGL/src/RenderDeviceGLImpl.cpp
+++ b/Graphics/GraphicsEngineOpenGL/src/RenderDeviceGLImpl.cpp
@@ -281,7 +281,9 @@ RenderDeviceGLImpl::RenderDeviceGLImpl(IReferenceCounters*       pRefCounters,
         m_DeviceInfo.NDC = NDCAttribs{-1.0f, 0.5f, 0.5f};
     }
 
-    if (m_GLCaps.FramebufferSRGB)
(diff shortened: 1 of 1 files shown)


TheMostDiligent:
The issue is more complex than this change because this is a global setting. Your change, in particular, will make all out-of-screen rendering incorrect as it will disable sRGB conversion when rendering to sRGB textures.

Admenri:
Test screenshot:
 Default:
unorm nochange
 Applying:
unorm afterchange

TheMostDiligent:
Your fix is incorrect as it globally disables the `FRAMEBUFFER_SRGB` capability when the swap chain does not use sRGB format. `FRAMEBUFFER_SRGB` is a remnant of OpenGL past and it must always be enabled to be consistent with other APIs.
In your specific case it only works because you use rgba8_unorm render target. If you use rgba8_unorm_srgb, which you should really do, it will break.
A correct fix is to only disable the sRGB for default framebuffer, not globally - I am looking into this.

Admenri:
You're right—this issue is far more complex than initially anticipated.
I'll raise this as a formal issue later for further investigation.

TheMostDiligent:
There is already an issue for this: https://github.com/DiligentGraphics/DiligentCore/issues/124

PR #691, opened 2025-08-02
Fix an issue that SwapChainDesc.DepthBufferFormat = TEX_FORMAT_R24G8_TYPELESS causing D3D12 invalid param error

hzqst:
An invalid param error `D3D12_CLEAR_VALUE::Format cannot be a typeless format. A fully qualified format must be supplied. Format = R24G8_TYPELESS.` was thrown when `ClearValue.Format` set to `DXGI_FORMAT_R24G8_TYPELESS`.

We just leave the `Desc.ClearValue.Format` empty and `GetClearFormat` wil return a proper format for us.

```cpp
//TextureD3D12Impl.cpp~L169
if (d3d12TexDesc.Flags & (D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET | D3D12_RESOURCE_FLAG_ALLOW_DEPTH_STENCIL))
    {
        if (m_Desc.ClearValue.Format != TEX_FORMAT_UNKNOWN)
            ClearValue.Format = TexFormatToDXGI_Format(m_Desc.ClearValue.Format);
        else
        {
            DXGI_FORMAT Format = TexFormatToDXGI_Format(m_Desc.Format, m_Desc.BindFlags);
            ClearValue.Format  = GetClearFormat(Format, d3d12TexDesc.Flags);
        }
```

### DIFF
--- a/Graphics/GraphicsEngineD3DBase/src/DXGITypeConversions.cpp
+++ b/Graphics/GraphicsEngineD3DBase/src/DXGITypeConversions.cpp
@@ -246,6 +246,12 @@ DXGI_FORMAT CorrectDXGIFormat(DXGI_FORMAT DXGIFormat, Uint32 BindFlags)
                 DXGIFormat = DXGI_FORMAT_D16_UNORM;
                 break;
 
+            case DXGI_FORMAT_R32G8X24_TYPELESS:
+            case DXGI_FORMAT_R32_FLOAT_X8X24_TYPELESS:
+            case DXGI_FORMAT_X32_TYPELESS_G8X24_UINT:
+                DXGIFormat = DXGI_FORMAT_D32_FLOAT_S8X24_UINT;
+                break;
+
             default:
                 break;
         }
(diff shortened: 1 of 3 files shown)


Graphics/GraphicsEngineD3D12/src/SwapChainD3D12Impl.cpp
```diff
@@ -112,7 +112,7 @@ void SwapChainD3D12Impl::InitBuffersAndViews()
         DepthBufferDesc.Usage       = USAGE_DEFAULT;
         DepthBufferDesc.BindFlags   = BIND_DEPTH_STENCIL;
 
-        DepthBufferDesc.ClearValue.Format               = DepthBufferDesc.Format;
+        DepthBufferDesc.ClearValue.Format               = TEX_FORMAT_UNKNOWN;
```
TheMostDiligent: I would suggest using `CorrectDXGIFormat(DepthBufferDesc.Format, BIND_DEPTH_STENCIL)` instead

TheMostDiligent (changes requested):

Graphics/GraphicsEngineD3D12/src/SwapChainD3D12Impl.cpp
hzqst: > CorrectDXGIFormat(DepthBufferDesc.Format, BIND_DEPTH_STENCIL)

image

It would be redundant with `CorrectDXGIFormat` stuffs, as `ClearValue.Format` takes `TEXTURE_FORMAT` instead of `DXGI_FORMAT`. I'm good with this though.

TheMostDiligent: Ah, OK.
So the right function here should be 
`GetDefaultTextureViewFormat(DepthBufferDesc.Format, TEXTURE_VIEW_DEPTH_STENCIL, DepthBufferDesc.BindFlags)`

PR #694, opened 2025-08-13
Ability to pass down DxCompilerPath to RenderStateCache

AlexanderDzhoganov:
Hi,

Here's a tiny change to add the option to pass the path to DXC in `RenderStateCacheCreateInfo`. Useful on Linux and OSX, otherwise it defaults to `dxcompiler` or `spv_dxcompiler`.

On a note, we have been running `DXCompilerLibraryLinux.cpp` on OSX with the `dxcompiler.dylib` shipped in the Vulkan SDK. I understand this is not supported from looking at the CMakeLists.txt but is there a technical reason or just lack of resources?

### DIFF
--- a/Graphics/GraphicsTools/interface/RenderStateCache.h
+++ b/Graphics/GraphicsTools/interface/RenderStateCache.h
@@ -76,6 +76,9 @@ struct RenderStateCacheCreateInfo
     /// shaders. If null, original source factory will be used.
     IShaderSourceInputStreamFactory* pReloadSource DEFAULT_INITIALIZER(nullptr);
 
+    /// DX Compiler path
+    const Char* DxCompilerPath DEFAULT_INITIALIZER(nullptr);
+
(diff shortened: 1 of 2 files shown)


TheMostDiligent:
> Here's a tiny change to add the option to pass the path to DXC in `RenderStateCacheCreateInfo`

I think a better way is to use the compiler path passed to the render device through the create info. This way there is no need for extra parameter and the compiler will be consistent with what the device uses.


> I understand this is not supported from looking at the CMakeLists.txt but is there a technical reason or just lack of resources?

This has not been needed/requested so far. Looks like it is not too difficult to add.

TheMostDiligent:
With [this change](https://github.com/DiligentGraphics/DiligentCore/commit/5ca848d5a319021eb85a2cd078ecde96725d7151), the cache will use the DXC path from the render device.

AlexanderDzhoganov:
Perfect, this makes more sense. Thank you!

PR #697, opened 2025-08-14
FIX: Clang and MSVC compile options duplicate for clang-cl

JKot-Coder:
### Summary

When building with clang-cl (Clang frontend for MSVC):

- Previously, both Clang-specific and MSVC-specific compiler flags were
  applied, producing duplicate warnings and huge warning output.
- Compilation could slow down by orders of magnitude due to -Wall/-Wextra.

### Changes

- Clang-cl now only uses MSVC-specific compile options.
- Pure Clang builds continue using Clang-specific warnings and suppressions.
- Redundant -Wall/-Wextra removed for clang-cl.

### Benefits

- Builds with clang-cl are significantly faster.
- Warning output is reasonable and manageable.
- Maintains correct compile flags for each compiler.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -468,14 +468,17 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
         message("Additional Clang compile options for release configurations: " ${DILIGENT_CLANG_RELEASE_COMPILE_OPTIONS})
     endif()
 
-    target_compile_options(Diligent-BuildSettings INTERFACE
-        ${DILIGENT_CLANG_COMPILE_OPTIONS}
-        # Some extra warnings
-        -Wall -Wextra -Wuninitialized -Wconditional-uninitialized -Wextra-tokens -Wpointer-arith -Wloop-analysis -Wunused
-        # Disable few warnings
-        -Wno-overloaded-virtual -Wno-incompatible-pointer-types-discards-qualifiers -Wno-unknown-pragmas
-        -Wno-zero-as-null-pointer-constant -Wno-unused-parameter -Wno-invalid-offsetof
-    )
+    target_compile_options(Diligent-BuildSettings INTERFACE ${DILIGENT_CLANG_COMPILE_OPTIONS})
+
+    if(NOT MSVC)
+        target_compile_options(Diligent-BuildSettings INTERFACE
+            # Some extra warnings
+            -Wall -Wextra -Wuninitialized -Wconditional-uninitialized -Wextra-tokens -Wpointer-arith -Wloop-analysis -Wunused
+            # Disable few warnings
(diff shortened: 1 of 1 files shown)


TheMostDiligent:
There are a few other places where clang-specific compiler options are set e.g.
Should they also be disabled for `clang-cl`?

https://github.com/DiligentGraphics/DiligentCore/blob/c4a98bd5d210d5ed093aee0f4715ead530466acf/Graphics/ShaderTools/CMakeLists.txt#L141-L143

https://github.com/DiligentGraphics/DiligentCore/blob/c4a98bd5d210d5ed093aee0f4715ead530466acf/Graphics/ShaderTools/CMakeLists.txt#L146-L149

https://github.com/DiligentGraphics/DiligentCore/blob/c4a98bd5d210d5ed093aee0f4715ead530466acf/Graphics/GraphicsEngineOpenGL/CMakeLists.txt#L223-L248

https://github.com/DiligentGraphics/DiligentCore/blob/c4a98bd5d210d5ed093aee0f4715ead530466acf/Tests/DiligentCoreTest/CMakeLists.txt#L25-L32

https://github.com/DiligentGraphics/DiligentCore/blob/c4a98bd5d210d5ed093aee0f4715ead530466acf/Tests/GPUTestFramework/CMakeLists.txt#L126-L141

https://github.com/DiligentGraphics/DiligentCore/blob/c4a98bd5d210d5ed093aee0f4715ead530466acf/Tests/DiligentCoreAPITest/CMakeLists.txt#L128-L141

JKot-Coder:
I'm 99% sure they all should be fixed same way. But I didn't checked any of this. 
I'm just started use Diligent and  faced with issued that clang++ fails to link at windows.
clang-cl - fails to compile 😄 So I'm just fixed most critical one issues for me. I'm use complied only core and dx12 for now.

TheMostDiligent:
Did you figure out how to make the build faster?
Last time I tried clang-cl, it took almost 1.5 hours to build everything.

TheMostDiligent:
This change does not seem to be correct.
If clang-cl only takes MSVC-like command line arguments, the entire branch should be disabled:

```
if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT MSVC)
```

This reduces the number of warnings, but build time is still ridiculous. It takes more than 5 minutes to build only `Diligent-GraphicsEngineD3D11-static`

JKot-Coder:
`if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT MSVC)`
Yes, that makes sense.

With all errors ignored using -w, the build completed in less than a minute. Warnings significantly affect build performance.

TheMostDiligent:
For me, it feels like linking takes most of the time.
Which generator do you use? VS or Ninja?

PR #705, opened 2025-09-03
making SelfDestroy virtual to fix #704

### LINKED ISSUE #704: MSVC optimization sucks with ReleaseWeakPtr virtual function call
As we knows `m_pRefCounters` can be allocated from another execution unit (Archiver dll in this case)

However MSVC compile the following virtual function call into direct function call:

```cpp
//RefCntAutoPtr.hpp

void Release() noexcept
{
    if (m_pRefCounters)
        m_pRefCounters->ReleaseWeakRef(); //will be optimized by MSVC 
    m_pRefCounters = nullptr;
    m_pObject      = nullptr;
}
```

This can lead to crash when my dll uses custom global operator_new override (mi_new to override `pNewRefCounters` allocation in this case) :

Image

```cpp
//RefCountedObjectImpl.hpp

pNewRefCounters = new RefCountersImpl{}; //this goes with `mi_new` from my dll and with ms crt `malloc` from archiver dll
```

### Possible fix A:

Image

Use custom allocator `m_pAllocator` to allocate `RefCountersImpl` when available.

### Possible fix B:

Tell MSVC not to optimize the virtual function call

```
#pragma optimize( "", off )
void Release() noexcept
{
    if (m_pRefCounters)
        m_pRefCounters->ReleaseWeakRef(); //will be optimized by MSVC 
    m_pRefCounters = nullptr;
    m_pObject      = nullptr;
}
#pragma optimize( "", on )
```

hzqst:
For https://github.com/DiligentGraphics/DiligentCore/issues/704

image

### DIFF
--- a/Common/interface/RefCountedObjectImpl.hpp
+++ b/Common/interface/RefCountedObjectImpl.hpp
@@ -46,8 +46,6 @@ namespace Diligent
 class RefCountersImpl : public IReferenceCounters
 {
 public:
-    virtual ~RefCountersImpl() {}
-
     inline virtual ReferenceCounterValueType AddStrongRef() override final
     {
         VERIFY(m_ObjectState.load() == ObjectState::Alive, "Attempting to increment strong reference counter for a destroyed or not initialized object!");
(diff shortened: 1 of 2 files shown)


TheMostDiligent:
I think that `RefCountersAllocationImpl` is not needed.
Just making `SelfDestroy` virtual should be enough.

PR #707, opened 2025-09-08
OpenXRUtilities fixes

n0bra1n3r:
- modified an include path in OpenXRUtilities.h to be the correct relative path.
- fixed a typo in the C API in OpenXRUtilities.cpp


n0bra1n3r:
Will remove merge commits and fix formatting.

PR #718, opened 2025-10-11
SPIRVResources: fix storage buffer name detection for shaders compiled by slang

hzqst:
as title.

### DIFF
--- a/Graphics/ShaderTools/src/SPIRVShaderResources.cpp
+++ b/Graphics/ShaderTools/src/SPIRVShaderResources.cpp
@@ -225,8 +225,8 @@ spv::ExecutionModel ShaderTypeToSpvExecutionModel(SHADER_TYPE ShaderType)
 }
 
 const std::string& GetUBOrSBName(diligent_spirv_cross::Compiler&               Compiler,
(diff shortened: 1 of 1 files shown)


Graphics/ShaderTools/src/SPIRVShaderResources.cpp
```diff
@@ -262,6 +262,19 @@ const std::string& GetUBName(diligent_spirv_cross::Compiler&               Compi
...
+{
+    //layout(std430, binding = 3) readonly buffer StructuredBuffer_world_material_t_std430_t_0 // <-- type name
+    //{
+    //    world_material_t_std430_0 _data[];
+    //} WorldMaterialSSBO; // <-- instance name
+
+    const std::string& instance_name = Compiler.get_name(SSB.id);
+    return ((IRSource.hlsl || IRSource.lang == spv::SourceLanguageSlang) && !instance_name.empty()) ? instance_name : SSB.name;
+}
+
```
TheMostDiligent: There is no need to have two identical functions with different names.
Since the logic is the exactly the same, `GetUBName` can be renamed to `GetUBOrSBName` and used for both.

PR #358, opened 2022-02-23
Platforms: Added recusive create and destroy directory

MikhailGorobets:

PR #726, opened 2025-11-26
Wayland support (vk only)

Lokomojo:
Re: [#358](https://github.com/DiligentGraphics/DiligentEngine/issues/358)

Turns out that only a minimal change is needed to get Wayland support working in diligent - at least for vulkan only.

This patch will only try to create a wayland window if NativeWindow::pWaylandSurface is defined, otherwise will fall back to xcb/xlib

### DIFF
--- a/BuildTools/CMake/BuildUtils.cmake
+++ b/BuildTools/CMake/BuildUtils.cmake
@@ -1,27 +1,8 @@
 if(PLATFORM_WIN32)
 
-    # Copies required dlls to the target's output directory
-    #
-    # The following dlls are copied:
-    #  - Engine dlls (GraphicsEngine*.dll)
-    #  - Archiver dll (Archiver*.dll)
-    #  - D3Dcompiler_47.dll
-    #  - Optional DXC dlls (dxcompiler.dll, dxil.dll, spv_dxcompiler.dll)
-    #
-    # Arguments:
-    #   TARGET_NAME  - name of the target to copy dlls for
-    #   DXC_REQUIRED - Indicates that the target requires DXC compiler dlls
-    #                  (dxcompiler.dll, dxil.dll and spv_dxcompiler.dll)
-    #
-    # Example:
-    #   copy_required_dlls(MyTarget DXC_REQUIRED YES)
-    #
-    function(copy_required_dlls TARGET_NAME)
-        set(options)
-        set(oneValueArgs DXC_REQUIRED)
-        set(multiValueArgs)
-        cmake_parse_arguments(PARSE_ARGV 1 arg "${options}" "${oneValueArgs}" "${multiValueArgs}")
-
+    # Copies engine dlls to the target's output directory
+    function(copy_engine_dlls TARGET_NAME)
+        set(ENGINE_DLLS)
         if(D3D11_SUPPORTED)
             list(APPEND ENGINE_DLLS Diligent-GraphicsEngineD3D11-shared)
         endif()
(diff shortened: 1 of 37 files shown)


TheMostDiligent:
I am not fully following the logic - are you trying to create surface for everything?
The expectation is that the user will try to initialize the swap chain for one specific surface, not all of them, so it looks like all is needed here is to add the Wayland surface and then properly use it.

Lokomojo:
Right yes we can get rid of the `foundPlatform` flag. I was thinking automatic fallback but it makes no sense since each windowing system has its own property in LinuxNativeWindow.

Graphics/GraphicsEngineVulkan/src/SwapChainVkImpl.cpp
```diff
@@ -102,17 +102,17 @@ void SwapChainVkImpl::CreateSurface()
 
         err = vkCreateMetalSurfaceEXT(m_Instance->GetVkInstance(), &surfaceCreateInfo, NULL, &m_VkSurface);
     }
-#elif defined(VK_USE_PLATFORM_WAYLAND_KHR)
-    if (m_Window.pDisplay != nullptr)
+#elif defined(VK_USE_PLATFORM_WAYLAND_KHR) || defined(VK_USE_PLATFORM_XCB_KHR) || defined(VK_USE_PLATFORM_XLIB_KHR)
```
TheMostDiligent, line 105: Empty branch that remained from your first version

remi-scarlet-1:
If I understood correctly the issue is merely Diligent's API missing the option to be initialized using a Wayland handle instead of an X11 one. Solving this seems rather simple, I wonder why this hasn't made it's way into the main branch yet. 

Am I missing something? Is there some philosophy to Diligent's API that would prevent this from being merged?
Would we have to implement Wayland support for the other graphics engine implementations too before this is merged?

TheMostDiligent:
This specific change has not been merged primarily because it fails to pass PR checks.
@remi-scarlet-1 Can you test this branch and see if it works for you?

remi-scarlet-1:
I'm currently setting up a project to test this. I'll hopefully have news to share tomorrow.

remi-scarlet-1:
Alright! As promised @TheMostDiligent, I gave this fork a spin and I’m happy to report it works, but with a caveat that likely explains why the CI tests failed.

The code changes implemented by @Lokomojo are correct, but they are currently being excluded from the build because `VK_USE_PLATFORM_WAYLAND_KHR` remains undefined. You can see the block here: 
[DiligentCore/Graphics/GraphicsEngineVulkan/src/SwapChainVkImpl.cpp#L106](https://github.com/Lokomojo/DiligentCore/blob/master/Graphics/GraphicsEngineVulkan/src/SwapChainVkImpl.cpp#L106)

I managed to get it working perfectly by forcing `VK_USE_PLATFORM_WAYLAND_KHR` in my CMake root. I assume the proper fix would be to handle this definition inside Diligent-BuildSettings (or the core CMake config), but since this is my first time debugging someone else's library's internals, I'm a bit too uncomfortable with the build system to make that change myself properly.

I hope this helps move things forward!

TheMostDiligent:
@remi-scarlet-1 thanks for confirming - I will fix and merge this PR.

Qubrick:
@TheMostDiligent 
Can you please add wayland for opengl as well? 
I think it use EGL_PLATFORM_WAYLAND_EXT

TheMostDiligent:
Implemented in aa774320d901a3845e94bdc1e074dda4f8ee296f and d060569afe4e1322b29a7f5cb91b693ba5eef51d
@Qubrick for OpenGL, can you submit a PR?

PR #733, opened 2025-12-12
Add support for push constants to SPIRVShaderResources

hzqst:
- [x] Add support for push constants to SPIRVShaderResources
- [x] Add tests that will compile a simple shader with each resource type, create SPIRVShaderResources from byte code and test that the expected resource is properly reflected.


Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
```diff
@@ -0,0 +1,336 @@
...
+        case SPIRVShaderResourceAttribs::ResourceType::AccelerationStructure:
+            return "AccelerationStructure";
+        case SPIRVShaderResourceAttribs::ResourceType::PushConstant:
+            return "PushConstant";
+        default:
+            return "Unknown";
+    }
+}
+
+std::vector<uint32_t> HLSLtoSPIRV(const char* FilePath, SHADER_TYPE ShaderType = SHADER_TYPE_PIXEL)
```
TheMostDiligent: This function can be used directly from ShaderTools, there is no need to make a copy here.

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
```diff
@@ -0,0 +1,336 @@
...
+    ShaderCI.pShaderSourceStreamFactory = pShaderSourceStreamFactory;
+
+    GLSLangUtils::InitializeGlslang();
+    auto SPIRV = GLSLangUtils::HLSLtoSPIRV(ShaderCI, GLSLangUtils::SpirvVersion::Vk100, nullptr, nullptr);
+    GLSLangUtils::FinalizeGlslang();
+
+    return SPIRV;
+}
+
+std::vector<uint32_t> GLSLtoSPIRV(const char* FilePath, SHADER_TYPE ShaderType = SHADER_TYPE_PIXEL)
```
TheMostDiligent: Same here, `GLSLtoSPIRV` can be used directly from ShaderTools

Graphics/ShaderTools/src/SPIRVShaderResources.cpp
```diff
@@ -145,6 +150,27 @@ SPIRVShaderResourceAttribs::SPIRVShaderResourceAttribs(const char*  _Name,
 // clang-format on
 {}
 
+// Test-only constructor for creating reference resources
+SPIRVShaderResourceAttribs::SPIRVShaderResourceAttribs(const char*        _Name,
```
TheMostDiligent: Since this is test-only, it it better not to have it in the public class API.
It is better to add a test-local struct, something like RefSPIRVResourceAttribs. It can be aggregate-initialized, which will be even simpler.

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
```diff
@@ -0,0 +1,336 @@
...
+#include "TestingEnvironment.hpp"
+#include "gtest/gtest.h"
+
+using namespace Diligent;
+using namespace Diligent::Testing;
+
+namespace
+{
+
+const char* SPIRVShaderResourceTypeToString(SPIRVShaderResourceAttribs::ResourceType Type)
```
TheMostDiligent: This one, on the other hand, seems to be potentially useful.
I would suggest moving it to SPIRVShaderResourceAttribs, make it static member `ResourceTypeToString`

TheMostDiligent (changes requested):

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
hzqst: This is not only just a copy but a workaround to load raw SPIRV when `DILIGENT_NO_HLSL` is defined (shader compiler not available at that time).

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
hzqst: workaround to load raw SPIRV when `DILIGENT_NO_GLSL` is defined (GLSLang not available at that time).

TheMostDiligent: Workarounds are not needed: if `DILIGENT_NO_GLSL`  or `NO_GLSLANG` is defined, the test should be disabled

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
```diff
@@ -0,0 +1,400 @@
...
+using namespace Diligent::Testing;
+
+namespace
+{
+
+struct SPIRVShaderResourceRefAttribs
+{
+    // clang-format off
+
+/*  0  */const char* const                                  Name;
```
TheMostDiligent: Bit fields are not needed here - they are for efficient packing.
For test, it is clearer and more concise to not use them.

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
```diff
@@ -0,0 +1,400 @@
...
+/* 20 */const Uint32                                        BufferStaticSize;
+/* 24 */const Uint32                                        BufferStride;
+/* 28 */
+/* 32 */ // End of structure
+
+    // clang-format on
+    //
+    // Test-only constructor for creating reference resources
+    SPIRVShaderResourceRefAttribs(const char*                              _Name,
+                                  SPIRVShaderResourceAttribs::ResourceType _Type,
```
TheMostDiligent: This constructor should be removed - the struct can be aggregate-initialized.
This will be clearer and shorter

Graphics/ShaderTools/include/SPIRVShaderResources.hpp
```diff
@@ -111,6 +113,20 @@ struct SPIRVShaderResourceAttribs
...
+                               Uint32       _BufferStaticSize) noexcept;
+
+    // Test-only constructor for creating reference resources
+    SPIRVShaderResourceAttribs(const char*        _Name,
+                               ResourceType       _Type,
+                               Uint16             _ArraySize,
+                               RESOURCE_DIMENSION _ResourceDim,
+                               Uint8              _IsMS,
+                               Uint32             _BufferStaticSize,
+                               Uint32             _BufferStride) noexcept;
```
TheMostDiligent: This constructor should not be needed.

Graphics/ShaderTools/include/SPIRVShaderResources.hpp
```diff
@@ -125,6 +141,39 @@ struct SPIRVShaderResourceAttribs
...
         return IsMS != 0;
     }
+
+    Uint32 GetConstantBufferSize() const
+    {
+        VERIFY((Type == PushConstant || Type == UniformBuffer), "Invalid resource type: PushConstant or UniformBuffer expected");
+        return BufferStaticSize;
+    }
+
+    Uint32 GetInlineConstantCountOrThrow(const char* ShaderName) const
```
TheMostDiligent, line 152: Looks like this function is not tested

Graphics/ShaderTools/src/SPIRVShaderResources.cpp
```diff
@@ -126,10 +131,50 @@ SPIRVShaderResourceAttribs::SPIRVShaderResourceAttribs(const diligent_spirv_cros
...
+    IsMS                          {0},
+    BindingDecorationOffset       {0}, // Push constants have no binding decoration
+    DescriptorSetDecorationOffset {0}, // Push constants have no descriptor set decoration
+    BufferStaticSize              {_BufferStaticSize},
+    BufferStride                  {0}
+// clang-format on
+{}
+
+// Test-only constructor for creating reference resources
+SPIRVShaderResourceAttribs::SPIRVShaderResourceAttribs(const char*        _Name,
```
TheMostDiligent: This constructor should be deleted

Tests/DiligentCoreTest/CMakeLists.txt
```diff
@@ -20,6 +20,12 @@ if(NOT WEBGPU_SUPPORTED)
     )
 endif()
 
+if(NOT VULKAN_SUPPORTED)
```
TheMostDiligent: Also, should be removed is GLSLang is not supported.
Can be tested as `NOT TARGET glslang`

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
```diff
@@ -0,0 +1,332 @@
...
+{
+    const char* const                              Name;
+    const Uint16                                   ArraySize;
+    const SPIRVShaderResourceAttribs::ResourceType Type;
+    const Uint8                                    ResourceDim; // RESOURCE_DIMENSION
+    const Uint8                                    IsMS;
+
+    // Offset in SPIRV words (uint32_t) of binding & descriptor set decorations in SPIRV binary
+    const uint32_t BindingDecorationOffset;
+    const uint32_t DescriptorSetDecorationOffset;
```
TheMostDiligent: These fields are not set/tested, they can be removed.

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+    const Uint32 BufferStaticSize;
+    const Uint32 BufferStride;
+};
+
+std::vector<unsigned int> LoadSPIRVFromHLSL(const char* FilePath, SHADER_TYPE ShaderType = SHADER_TYPE_PIXEL)
```
TheMostDiligent, line 73: This function can be used from ShaderTools. If GLSLang is disabled, the test can also be disabled - this is not a problem.

hzqst: I don't get it, dym "we should move LoadSPIRVFromHLSL to ShaderTools" ?, or "There is a function doing identical job in ShaderTools"?

TheMostDiligent: Ah, sorry, did not read this properly.
You are already using them.
There is no need to use InitializeGlslang/FinalizeGlslang.
It is called by [TestingEnvironmentVk](https://github.com/DiligentGraphics/DiligentCore/blob/b87073820b2ce4d9bcb99b67f744cc50aca67142/Tests/GPUTestFramework/src/Vulkan/TestingEnvironmentVk.cpp#L47)

PR #734, opened 2025-12-14
Add DXC tests to SPIRVShaderResourcesTest, based on #728

hzqst:
- [x] Add DXC tests to SPIRVShaderResourcesTest.

### DIFF
--- a/Tests/DiligentCoreTest/assets/shaders/SPIRV/InputAttachments.psh
+++ b/Tests/DiligentCoreTest/assets/shaders/SPIRV/InputAttachments.psh
@@ -1,0 +1,14 @@
+// Test for input attachments (subpass inputs)
+// In HLSL, subpass inputs are declared using SubpassInput type with vk::input_attachment_index attribute
+[[vk::input_attachment_index(0)]]
+[[vk::binding(0)]]
+SubpassInput<float4> g_InputAttachment;
+
+float4 main() : SV_Target
+{
+    // Read from input attachment (subpass input)
+    float4 color = g_InputAttachment.SubpassLoad();
+
+    // Process the input color
+    return color * 2.0;
+}
--- a/Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
+++ b/Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
@@ -1,0 +1,320 @@
+/*
+ *  Copyright 2025 Diligent Graphics LLC
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *  In no event and under no legal theory, whether in tort (including negligence),
+ *  contract, or otherwise, unless required by applicable law (such as deliberate
+ *  and grossly negligent acts) or agreed to in writing, shall any Contributor be
+ *  liable for any damages, including any direct, indirect, special, incidental,
+ *  or consequential damages of any character arising as a result of this License or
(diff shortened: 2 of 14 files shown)


Graphics/ShaderTools/CMakeLists.txt
```diff
@@ -227,7 +238,7 @@ add_custom_command(OUTPUT ${HLSL_DEFINITIONS_INC} # We must use full path here!
                    VERBATIM
 )
 
-set_common_target_properties(Diligent-ShaderTools)
+set_common_target_properties(Diligent-ShaderTools 17)
```
TheMostDiligent: This is not needed, c++ 17 is the default now

Tests/DiligentCoreTest/assets/shaders/SPIRV/InputAttachments.psh
```diff
@@ -11,4 +11,4 @@ float4 main() : SV_Target
 
     // Process the input color
     return color * 2.0;
-}
+}
```
TheMostDiligent: Does your editor remove new lines at the end of the file?

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
```diff
@@ -43,18 +44,23 @@ using namespace Diligent::Testing;
 
 namespace
 {
+std::unique_ptr<IDXCompiler> g_pDXCompiler;
```
TheMostDiligent: It is better to make this static member of a the test class

TheMostDiligent (changes requested):
Can you please submit these two changes as separate PRs?
It is much easier to review changes when they are not related.

Graphics/ShaderTools/CMakeLists.txt
hzqst: ShaderTools restored. gonna push in a separate pr.

hzqst:
> Can you please submit these two changes as separate PRs?

I will, once this pr get merged.

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
```diff
@@ -184,13 +229,47 @@ void TestSPIRVResources(const char*                                       FilePa
     }
 }
 
+void TestSPIRVResources(const char*                                       FilePath,
+                        const std::vector<SPIRVShaderResourceRefAttribs>& RefResources,
+                        SHADER_TYPE                                       ShaderType            = SHADER_TYPE_PIXEL,
+                        const char*                                       CombinedSamplerSuffix = nullptr,
```
TheMostDiligent: `CombinedSamplerSuffix` is never used from what I can tell. It can be removed in this case.

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
```diff
@@ -184,13 +229,47 @@ void TestSPIRVResources(const char*                                       FilePa
...
+                        bool                                              IsGLSL                = false)
```
TheMostDiligent: I suggest use `SHADER_SOURCE_LANGUAGE SourceLanguage` parameter for clarity

TheMostDiligent (changes requested):
I think I did not explain my intent clear: DXC tests should be executed along with glslang, which is still the primary compiler for HLSL.
So each HLSL shader should be tested with glslang _and_ DXC

hzqst:
> So each HLSL shader should be tested with glslang _and_ DXC


I believe this is "HLSL tested with glslang and DXC" or I get it wrong ?

image

TheMostDiligent:
Can you please update the PR.
I updated the branch and now GitHub shows very large diff.
A couple of comments:
- I separated the `copy_required_dlls` CMake function into a few helper functions. There is now the [`copy_compiler_dlls`](https://github.com/DiligentGraphics/DiligentCore/blob/f8969699e037d3bd3447960ed2d3e7d065915df9/BuildTools/CMake/BuildUtils.cmake#L46) function that can be used to only copy the `spv_dxcompiler.dll`:
`copy_shader_compiler_dlls(DiligentCoreTest DXCOMPILER_FOR_SPIRV YES)`
- Can you move DXC compiler tests into their own, so that for each shader there is a pair of tests, one for glslang, one for DXC. It is  preferred that unit tests test just one single thing. In this case, glslang and DXC use different paths and it is better to have them separate

PR #741, opened 2025-12-24
Native vulkan draw tests for ConvertUBOToPushConstants

hzqst:


hzqst:
@TheMostDiligent 

```cpp
 // Step 4: Create renderer with patched shaders
    PatchedPushConstantsRenderer Renderer{
        pSwapChain,
        pTestingSwapChainVk->GetRenderPass(),
        VS_SPIRV,
        FS_SPIRV_Patched,
        sizeof(PushConstantData),
        VK_SHADER_STAGE_FRAGMENT_BIT};

    // Step 5: Render with push constants
    VkCommandBuffer vkCmdBuffer = pEnv->AllocateCommandBuffer();

    pTestingSwapChainVk->BeginRenderPass(vkCmdBuffer,
                                         VK_PIPELINE_STAGE_VERTEX_SHADER_BIT | VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT,
                                         ClearColor);

    // Set push constant data - Factor = (1,1,1,1) to make output identical to reference
    PushConstantData PushData = {};
    PushData.Factor[0]        = 1.0f;
    PushData.Factor[1]        = 1.0f;
    PushData.Factor[2]        = 1.0f;
    PushData.Factor[3]        = 1.0f;

    Renderer.Draw(vkCmdBuffer, &PushData, sizeof(PushData));

    pTestingSwapChainVk->EndRenderPass(vkCmdBuffer);
    vkEndCommandBuffer(vkCmdBuffer);
    pEnv->SubmitCommandBuffer(vkCmdBuffer, true);

    // Step 6: Comparison native draw image with ref snapshot
    pTestingSwapChainVk->NativeDrawCompareWithSnapshot(nullptr); //<--- TakeSnapshot(nullptr, true) ?
```

The `NativeDrawCompareWithSnapshot` is just another `TakeSnapshot` but comparing mapped data with ReferenceImage instead of saving mapped data to ReferenceImage. what about TakeSnapshot(ITexture *, **bool CompareToRefImage**) ?

or what about :

```
pTestingSwapChainVk->TakeSnapshot(ITexture *, [](){
    CompareWithReferenceImage(...
});
```

TheMostDiligent:
I think the way it should be done is to add the normal test (like draw command test), use testing swap chain to capture the reference, but then implement all the logic in native Vulkan. This should not require changing any testing swap chain implementations.

hzqst:
> I think the way it should be done is to add the normal test (like draw command test), use testing swap chain to capture the reference, but then implement all the logic in native Vulkan. This should not require changing any testing swap chain implementations.

I've tried `TestingSwapChain->CompareWithSnapshot` but it was not working at all. nothing was found in the captured image. probably because the `TestingSwapChainVk::BeginRenderPass` uses `TestingSwapChainVk::m_vkFramebuffer` instead of `TestingSwapChainVk::m_pRenderTarget` as destination ?

-- nvm, done by rendering on top of `TestingSwapChainVk::m_pRenderTarget`

TheMostDiligent:
You can use the exact same test flow, render the reference and take snapshot. Then instead of using Diligent API (e.g. calling draw), get the Vulkan command list with IDeviceContextVk::GetVkCommandBuffer and proceed with native Vulkan. Then call present that will compare rendered images.

PR #742, opened 2025-12-24
ConvertUBOToPushConstantPass: better handle edge cases

TheMostDiligent:
A few fixes suggested by the Chat:

1. Return `SuccessWithoutChange` if UBO is not found, to be more consistent with SPIRV-Tools behavior. That should be fine for us, we can check if the UBO is present from `SPIRVShaderResources`.

2. Pointer SSA cycles through OpPhi can infinite-recursively traverse

You tried to guard cycles with seen, but the guard is only applied in the branch where the instruction already has push-constant storage class:

```
if (IsPointerToStorageClass(inst, PushConstant)) {
    if (inst->opcode() == OpPhi) { insert seen ... }
    recurse users...
    if (OpPhi) erase ...
}
```

In the other path (where you change the result type), OpPhi is handled in the switch:

```
case OpPhi:
    ChangeResultStorageClass(inst);
    recurse users...   // <-- no seen protection here
    return true;
```

If you have a classic pointer-phi cycle (very common in control flow graphs), you can end up in infinite recursion or deep repeated work.

Fix: apply the seen/cycle guard for OpPhi before recursing in both branches (or more generally, guard by result_id() for any pointer-producing instruction you revisit).

A simple pattern:

* if inst->result_id() != 0 and !seen->insert(inst->result_id()).second → stop
* then recurse
* (optionally) don’t erase; just let seen be “visited” rather than “in current stack”

Using “visited” instead of “stack” is usually enough and simpler.

3. PropagateStorageClass only covers some pointer-producing ops; you will hit real-world misses

You cover the common ones (access chains, copy, phi, select). But real shaders can include additional pointer-producing ops, depending on SPIR-V version/extensions:

* OpConstantNull can be a null pointer of some pointer type (rare but valid)
* OpUndef of a pointer type
* OpFunctionParameter (pointers passed into helper functions) — you currently won’t touch it (and your root is a global var, but it can feed into calls in some patterns)
* OpConvertPtrToU / OpConvertUToPtr (less common; extension-y)
* OpBitcast can produce a pointer result in some patterns; you excluded OpBitcast by comment (“result type independent”), but if the result is a pointer type, it absolutely matters.

You currently have:

```
case OpBitcast:
    return false;
```

If a pointer is bitcast (it happens), you’ll leave a pointer result typed as Uniform while its operand chain was changed to PushConstant, and validation will fail.

**Recommendation:** If IsPointerResultType(inst) is true and you see OpBitcast, you likely need to treat it like OpCopyObject/OpSelect and update its result type to push-constant pointer of the same pointee.

Also, instead of enumerating, you can follow the exact opcode set used by SPIRV-Tools’ own storage class fixing pass (you referenced it in comment). If you’re mirroring fix_storage_class.cpp, mirror it fully.

### DIFF
--- a/Graphics/ShaderTools/src/ConvertUBOToPushConstant.cpp
+++ b/Graphics/ShaderTools/src/ConvertUBOToPushConstant.cpp
@@ -200,6 +200,7 @@ public:
         spvtools::opt::Instruction* ptr_type_inst = get_def_use_mgr()->GetDef(target_var->type_id());
         if (ptr_type_inst == nullptr || ptr_type_inst->opcode() != spv::Op::OpTypePointer)
         {
+            LOG_ERROR_MESSAGE("Failed to convert UBO block '", m_BlockName, "': target variable has unexpected type.");
             return Status::Failure;
         }
 
@@ -214,6 +215,7 @@ public:
         if (new_ptr_type_id == 0)
         {
             // Failed to create new pointer type
+            LOG_ERROR_MESSAGE("Failed to convert UBO block '", m_BlockName, "': could not create PushConstant pointer type.");
             return Status::Failure;
         }
 


TheMostDiligent:
@hzqst I implemented a few fixes suggested by Chat, can you take a look and check them?

hzqst:
> @hzqst I implemented a few fixes suggested by Chat, can you take a look and check them?

all look good to me

PR #743, opened 2025-12-25
Fix ConvertUBOToPushConstant: The pointer type instruction need to be reordered

hzqst:
otherwise the SPIRV might be rejected by validation layer due to violating SPIR-V validation rules.

### DIFF
--- a/Graphics/ShaderTools/src/ConvertUBOToPushConstant.cpp
+++ b/Graphics/ShaderTools/src/ConvertUBOToPushConstant.cpp
@@ -280,8 +280,10 @@ private:
     bool ComesBeforeInTypesValues(const spvtools::opt::Instruction* a,
                                   const spvtools::opt::Instruction* b) const
     {
-        if (a == nullptr || b == nullptr || a == b)
+        if (a == nullptr || b == nullptr)
             return false;
+        if (a == b)
+            return true; // Same instruction, no reordering needed
 
         bool seen_a = false;
         for (auto& inst : context()->module()->types_values())


TheMostDiligent:
Review from Chat:

Main issue: your “new type was created” detection is wrong

This check:

```
if (new_ptr_type_inst != nullptr && new_ptr_type_inst != ptr_type_inst)
{
    // The new pointer type was created (not reused).
    ...
}
```

does not mean it was newly created.

It only means “the new pointer type instruction is not the old pointer type instruction”.

If FindPointerToType() returns an existing pointer type (reused), then new_ptr_type_inst will still differ from ptr_type_inst, and you’ll move an existing type instruction.

Why that’s dangerous

If that reused pointer type already had uses somewhere earlier, and you move it around, you can easily break the “defined before use” rule for those existing uses. In other words, this can introduce validation failures in shaders that previously validated.

Same problem in ChangeResultStorageClass().

**Minimal safe condition**

You need an ordering check: “is new_type_inst after the use site?”

For the first snippet, the use site is target_var:

* if new_ptr_type_inst already appears before target_var in types_values(), don’t touch it.
* if it appears after target_var, you should move it earlier.

Likewise in ChangeResultStorageClass(), compare new_type_inst against &inst (or against the first place you’ll use it).

Important: never move a type later

Your current code always does:

```
new_type_inst->InsertAfter(result_type_inst);
```

That could move an existing type forward (later) even if it was already in a valid spot. You want a rule like:

* Only move it if it is currently after the use
* Move it to a location before the use (in types section)

```
// Returns true if 'a' appears before 'b' in module()->types_values().
static bool ComesBeforeInTypesValues(spvtools::opt::Module* module,
                                    const spvtools::opt::Instruction* a,
                                    const spvtools::opt::Instruction* b)
{
    bool seen_a = false;
    for (auto& inst : module->types_values())
    {
        if (&inst == a)
        {
            seen_a = true;
        }
        else if (&inst == b)
        {
            return seen_a;
        }
    }
    // If either is not in types_values(), be conservative.
    return false;
}

// Ensures 'type_inst' is placed before 'use_inst' in types_values().
// Only moves when type_inst is currently after use_inst.
static void EnsureTypeBeforeUseInTypesValues(spvtools::opt::IRContext* ctx,
                                            spvtools::opt::Instruction* type_inst,
                                            spvtools::opt::Instruction* use_inst)
{
    if (type_inst == nullptr || use_inst == nullptr)
        return;

    auto* module = ctx->module();

    // If already before, do nothing.
    if (ComesBeforeInTypesValues(module, type_inst, use_inst))
        return;

    // Otherwise, move it immediately before the use.
    // (Both should be in types_values(): types/consts/global vars live there.)
    type_inst->RemoveFromList();
    type_inst->InsertBefore(use_inst);
}

```


```
auto* new_ptr_type_inst = get_def_use_mgr()->GetDef(new_ptr_type_id);
EnsureTypeBeforeUseInTypesValues(context(), new_ptr_type_inst, target_var);
```

```
auto* new_type_inst = get_def_use_mgr()->GetDef(new_result_type_id);
EnsureTypeBeforeUseInTypesValues(context(), new_type_inst, &inst);
```



After physically moving instructions around, it’s a good idea to invalidate analyses (depending on SPIRV-Tools version / how your pipeline runs passes):

```
context()->InvalidateAnalyses(spvtools::opt::IRContext::kAnalysisDefUse |
                              spvtools::opt::IRContext::kAnalysisTypes);
```

hzqst:
Fixed by introducing `EnsureTypeBeforeUseInTypesValues` -- compares the relative order of `type_inst` and the actual use site (`target_var` / `&inst`) inside `module()->types_values()`: it only moves the type definition when it currently appears *after* the use, and it only moves it earlier.

Graphics/ShaderTools/src/ConvertUBOToPushConstant.cpp
```diff
@@ -270,6 +275,74 @@ class ConvertUBOToPushConstantPass : public spvtools::opt::Pass
     }
 
 private:
+    // Returns true if instruction 'a' appears before instruction 'b' in module->types_values().
+    // This is used to check if a type definition comes before its use in the SPIR-V module.
+    bool ComesBeforeInTypesValues(const spvtools::opt::Instruction* a,
+                                  const spvtools::opt::Instruction* b) const
+    {
+        if (a == nullptr || b == nullptr || a == b)
```
TheMostDiligent: If a == b reordering should not be needed I suppose:

```
if (a == nullptr || b == nullptr)
    return false;
if (a == b)
    return true;
```

hzqst: fixed

PR #744, opened 2025-12-25
Added HLSL Compiler Profile Arguments To ShaderCreateInfo

zackory137c-hash:
I wanted to enable some features from DXC for HLSL compilation that isn't support directly by Diligent. This PR adds the `profile` field to ShaderCreateInfo to allow custom profile arguments passed into HLSL compilation.


TheMostDiligent:
Shader profile string is determined by the shader type and shader version, which are in create info.
What kind of DXC shader profile did you need?

PR #747, opened 2025-12-30
Full vulkan inlineconstants support, based on #729

hzqst:
- Extends `PipelineStateVkImpl::ShaderStageInfo` with `ShaderResources` so shader reflection data can be reconstructed/updated after SPIR-V patching (e.g., UBO → push constants).
- Introduces `PushConstantInfoVk` and stores push-constant metadata in `PipelineLayoutVk`; `PipelineLayoutVk::Create(...)` now configures `VkPipelineLayoutCreateInfo.pPushConstantRanges` (and validates size vs `maxPushConstantsSize`).
- Adds logic in Vulkan PSO creation to derive push-constant info from signatures and to patch shaders via `ConvertUBOToPushConstants`, updating both SPIR-V and the associated shader resources.
- Tweaks validation so push constants are validated but skip descriptor-set resource cache checks; fixes descriptor counting to use `ResDesc.GetArraySize()`.

- Adds `ShaderVkImpl::CreateSPIRVShaderResources(const std::vector<uint32_t>&)` to rebuild `SPIRVShaderResources` from SPIR-V after `ShaderVkImpl::Initialize`.
- Persists CI-derived state (`EntryPoint`, `LoadConstantBufferReflection`) so reconstructed resources match the original reflection configuration.
- Changes `GetShaderResources()` to return a null shared_ptr while compiling asynchronously instead of asserting.

- Updates `SerializedPipelineStateImpl::PatchShadersVk` to also transfer `ShaderResources` into `PipelineStateVkImpl::ShaderStageInfo`, ensuring archived/serialized Vulkan PSOs keep the per-stage reflection/resources alongside shaders and SPIR-V.

### DIFF
--- a/Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
+++ b/Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
@@ -827,11 +827,11 @@ bool PipelineStateVkImpl::InitPushConstantInfoFromSignatures(PushConstantInfoVk&
     // Vulkan allows only one push constant range per pipeline layout.
     // DiligentCore allows multiple inline constant resources, so we promote only the first inline constant
     // from resource signatures to push constants. Other inline constants remain uniform buffers.
-    const PipelineResourceDesc* pResDesc             = nullptr;
-    const char*                 PushConstantName     = nullptr;
-    Uint32                      PushConstantSize     = 0;
-    Uint32                      PushConstantSignIdx  = INVALID_PUSH_CONSTANT_INDEX;
-    Uint32                      PushConstantResIdx   = INVALID_PUSH_CONSTANT_INDEX;
+    const PipelineResourceDesc* pResDesc            = nullptr;
+    const char*                 PushConstantName    = nullptr;
+    Uint32                      PushConstantSize    = 0;
+    Uint32                      PushConstantSignIdx = INVALID_PUSH_CONSTANT_INDEX;
+    Uint32                      PushConstantResIdx  = INVALID_PUSH_CONSTANT_INDEX;
 
     for (Uint32 s = 0; s < m_SignatureCount; ++s)
     {


TheMostDiligent:
It is in fact hard to look at these changes separated from the context.
Can you submit the whole remaining changes as one PR - this will be easier to see why they are needed?

Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
```diff
@@ -784,6 +821,97 @@ void PipelineStateVkImpl::RemapOrVerifyShaderResources(
...
+                            }
+
+                            // Found push constant from SPIR-V
+                            return true;
+                        }
+                    }
+                }
+            }
+        }
+    }
```
TheMostDiligent, line 895: Why is this needed? Signature must be the only source of truth.
If it defines the resource as push constants, it will be patched as push constants.
If implicit resource signature is used, it will derive the push constants block from the shader.
In case the shader defines the block as push constants, but the signature defines it as a UBO, we will throw an error. (in theory, it should be possible to patch the other way, e.g. push constants -> UBO, but it is unnecessary and we will not support this).

Graphics/GraphicsEngineVulkan/include/ShaderVkImpl.hpp
```diff
@@ -92,7 +92,11 @@ class ShaderVkImpl final : public ShaderBase<EngineVkImplTraits>
 
     const std::shared_ptr<const SPIRVShaderResources>& GetShaderResources() const
     {
-        DEV_CHECK_ERR(!IsCompiling(), "Shader resources are not available until the shader is compiled. Use GetStatus() to check the shader status.");
+        static const std::shared_ptr<const SPIRVShaderResources> NullShaderResource;
+        // NOTE: while shader is compiled asynchronously, is not available
+        if (IsCompiling())
+            return NullShaderResource;
```
TheMostDiligent, line 98: This does not look right.
If the shader is compiling, no method should be called except for `GetStatus()`

Graphics/GraphicsEngineVulkan/include/DeviceContextVkImpl.hpp
```diff
@@ -60,6 +60,9 @@
 #include "HashUtils.hpp"
 #include "ManagedVulkanObject.hpp"
 
+// Push constants data storage (max size is typically 128-256 bytes, use 256 for safety)
+#define DILIGENT_MAX_PUSH_CONSTANTS_SIZE 256
```
TheMostDiligent, line 64: Let's move this constant to `Constants.h`, and call it `DILIGENT_MAX_INLINE_CONSTANTS_SIZE` / `MAX_INLINE_CONSTANTS_SIZE` similar to other constants.
It is already used in D3D11 and D3D12, so let's make it unified.
Can you submit this as separate PR?

Graphics/GraphicsEngineVulkan/include/VulkanUtilities/CommandBuffer.hpp
```diff
@@ -504,6 +504,18 @@ class CommandBuffer
...
+                                     VkShaderStageFlags stageFlags,
+                                     uint32_t           offset,
+                                     uint32_t           size,
+                                     const void*        pValues)
+    {
+        VERIFY_EXPR(m_VkCmdBuffer != VK_NULL_HANDLE);
+        VERIFY_EXPR(pValues != nullptr);
+        VERIFY_EXPR(size > 0);
+        vkCmdPushConstants(m_VkCmdBuffer, layout, stageFlags, offset, size, pValues);
+    }
```
TheMostDiligent, line 517: Can you also submit this small change as separate PR?
It will reduce the number of changed files

Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
hzqst: Fixed by: always promote only the first inline constant to push constants.

Graphics/GraphicsEngineVulkan/src/DeviceContextVkImpl.cpp
```diff
@@ -1075,11 +1206,18 @@ void DeviceContextVkImpl::PrepareForRayTracing()
     EnsureVkCmdBuffer();
 
     ResourceBindInfo& BindInfo = GetBindInfo(PIPELINE_TYPE_RAY_TRACING);
+
+    // Update inline constant buffers before binding descriptor sets
+    UpdateInlineConstantBuffers(BindInfo);
```
TheMostDiligent, line 1211: I think that the two-stage inline constant data caching logic (first in the resource cache and second in the device context) is overly complicated.
Inline constants are intended to be very small and changing with every draw call.
In D3D11 and D3D12, inline constants are committed before every draw and not tracked to be dirty.
I suggest Vulkan works similarly.
Tracking logic + copying from SRB to device context adds unnecessary overhead for the intended use case while helps with unintended (e.g. when inline constants don't change between draw calls).
If constants don't change every time, the resource should not really be an inline constant.

Graphics/GraphicsEngineVulkan/src/PipelineLayoutVk.cpp
```diff
@@ -59,7 +59,10 @@ void PipelineLayoutVk::Release(RenderDeviceVkImpl* pDeviceVk, Uint64 CommandQueu
     }
 }
 
-void PipelineLayoutVk::Create(RenderDeviceVkImpl* pDeviceVk, RefCntAutoPtr<PipelineResourceSignatureVkImpl> ppSignatures[], Uint32 SignatureCount) noexcept(false)
+void PipelineLayoutVk::Create(RenderDeviceVkImpl*                            pDeviceVk,
+                              RefCntAutoPtr<PipelineResourceSignatureVkImpl> ppSignatures[],
+                              Uint32                                         SignatureCount,
+                              const PushConstantInfoVk&                      PushConstant) noexcept(false)
```
TheMostDiligent, line 65: PipelineLayoutVk should be completely responsible for initializing the layout.
Right now the logic is split between `PipelineLayoutVk::Create` and `PipelineStateVkImpl::InitPushConstantInfoFromSignatures`, but everything should be in `PipelineLayoutVk`

Graphics/GraphicsEngineVulkan/src/ShaderVkImpl.cpp
```diff
@@ -229,47 +229,63 @@ void ShaderVkImpl::Initialize(const ShaderCreateInfo& ShaderCI,
...
         }
 
         m_Status.store(SHADER_STATUS_READY);
     }
 }
 
+std::shared_ptr<const SPIRVShaderResources> ShaderVkImpl::CreateSPIRVShaderResources(const std::vector<uint32_t>& SPIRV) noexcept(false)
+{
+    //This ensures m_EntryPoint always initialized from ShaderCI.EntryPoint when reconstructing shader resources with CreateSPIRVShaderResources
+    m_EntryPoint = m_CIEntryPoint;
```
TheMostDiligent, line 265: The fact that after calling CreateSPIRVShaderResources changes the is both very unclear and dangerous.
In particular, multiple PSO may use patch same shader - this is totally valid.
This will create a race condition here.

I reworked the SPIRVShaderResources to add a static create method that should be used here instead.

Graphics/GraphicsEngineVulkan/include/ShaderResourceCacheVk.hpp
```diff
@@ -287,16 +323,29 @@ class ShaderResourceCacheVk : public ShaderResourceCacheBase
 
     std::unique_ptr<void, STDDeleter<void, IMemoryAllocator>> m_pMemory;
 
+    // Memory for inline constant data (allocated separately, freed in destructor)
+    // All inline constants use this memory for CPU-side staging.
+    // Push constant selection is done at PSO level, not cache level.
+    std::unique_ptr<void, STDDeleter<void, IMemoryAllocator>> m_pInlineConstantMemory;
```
TheMostDiligent, line 329: Other implementations, D3D12 in particular, use the same chunk of memory for all data.
Would be nice to follow this patter here as well.

Graphics/GraphicsEngineVulkan/src/PipelineResourceSignatureVkImpl.cpp
```diff
@@ -474,6 +547,54 @@ void PipelineResourceSignatureVkImpl::CreateSetLayouts(const bool IsSerialized)
...
+                if (ResDesc.VarType == SHADER_RESOURCE_VARIABLE_TYPE_STATIC)
+                {
+                    const Uint32 DataSize    = InlineCBAttr.NumConstants * sizeof(Uint32);
+                    const Uint32 CacheOffset = Attr.StaticCacheOffset;
+                    m_pStaticResCache->InitializeInlineConstantBuffer(0, CacheOffset, InlineCBAttr.NumConstants, pCurrentDataPtr);
+                    pCurrentDataPtr += DataSize;
+                }
+            }
+        }
+    }
```
TheMostDiligent, line 597: This logic belongs to shader resource cache

Graphics/GraphicsEngineVulkan/src/PipelineResourceSignatureVkImpl.cpp
```diff
@@ -509,10 +645,60 @@ void PipelineResourceSignatureVkImpl::InitSRBResourceCache(ShaderResourceCacheVk
...
+        // Each SRB has its own copy of the staging data, but shares the GPU buffer
+        void* pInlineConstantMemory = nullptr;
+        if (TotalInlineConstantSize > 0)
+        {
+            pInlineConstantMemory = CacheMemAllocator.Allocate(TotalInlineConstantSize, "Inline constant data", __FILE__, __LINE__);
+            memset(pInlineConstantMemory, 0, TotalInlineConstantSize);
+
+            // Pass ownership of the memory to the resource cache for proper cleanup
+            ResourceCache.SetInlineConstantMemory(CacheMemAllocator, pInlineConstantMemory);
+        }
```
TheMostDiligent, line 684: This logic belongs to ShaderResourceCache

Graphics/GraphicsEngineVulkan/src/DeviceContextVkImpl.cpp
hzqst: Fixed in https://github.com/hzqst/DiligentCore/commit/c2ed69c6f60cdd9747734caa6b61016d6a3beb7b

Graphics/GraphicsEngineVulkan/src/PipelineLayoutVk.cpp
hzqst: Fixed in https://github.com/hzqst/DiligentCore/commit/c65c3abe378dc881de177d21858607ae8ad5be3a

Graphics/GraphicsEngineVulkan/src/ShaderVkImpl.cpp
hzqst: Fixed in https://github.com/hzqst/DiligentCore/commit/72ba7c94f58461fa9a05a68e1dad6705f7cf2e8b

Graphics/GraphicsEngineVulkan/include/ShaderResourceCacheVk.hpp
hzqst: Fixed in https://github.com/hzqst/DiligentCore/commit/2ccccaf8f633d11b0f171057a6ca5d27e38f7b75

Graphics/GraphicsEngineVulkan/src/PipelineResourceSignatureVkImpl.cpp
hzqst: Fixed in https://github.com/hzqst/DiligentCore/commit/2ccccaf8f633d11b0f171057a6ca5d27e38f7b75

PR #750, opened 2026-01-02
Final changes to enable inline constants in Vulkan

hzqst:


Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
```diff
@@ -784,8 +823,80 @@ void PipelineStateVkImpl::RemapOrVerifyShaderResources(
...
+        {
+            const ShaderVkImpl* pShader = Stage.Shaders[i];
+
+            if (!Stage.ShaderResources[i] && !pShader->IsCompiling())
+            {
+                SPIRVShaderResources::CreateInfo ResCI;
+                ResCI.ShaderType                  = pShader->GetDesc().ShaderType;
+                ResCI.Name                        = pShader->GetDesc().Name;
+                ResCI.CombinedSamplerSuffix       = pShader->GetDesc().UseCombinedTextureSamplers ? pShader->GetDesc().CombinedSamplerSuffix : nullptr;
+                ResCI.LoadShaderStageInputs       = pShader->GetDesc().ShaderType == SHADER_TYPE_VERTEX;
```
TheMostDiligent: Loading shader stage inputs is not needed.
It is used to patch Vertex Shader inputs, which is done when the shader is created.
They are already correct in the original SPIRV, and are the same in patched.

Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
```diff
@@ -784,8 +823,80 @@ void PipelineStateVkImpl::RemapOrVerifyShaderResources(
...
+                ResCI.LoadUniformBufferReflection = true; //LoadConstantBufferReflection;
```
TheMostDiligent: Why is uniform buffer reflection needed?

Graphics/GraphicsEngineVulkan/include/PipelineLayoutVk.hpp
```diff
@@ -70,6 +91,18 @@ class PipelineLayoutVk
...
+    Uint32 m_PushConstantSize = 0;
+
+    // Shader stages that use push constants
+    VkShaderStageFlags m_PushConstantStageFlags = 0;
+
+    // Index of the signature containing the push constant resource
+    Uint32 m_PushConstantSignatureIndex = INVALID_PUSH_CONSTANT_INDEX;
+
+    // Resource index within the signature for push constant
+    Uint32 m_PushConstantResourceIndex = INVALID_PUSH_CONSTANT_INDEX;
```
TheMostDiligent: I suggest grouping all these variables into internal struct `PushConstantInfo`:

```cpp
struct PushConstantInfo
{
    Uint32 Size = 0;
    VkShaderStageFlags StageFlags = 0;
    ...
    constexpr operator bool() { return Size != 0; }
};
```

This will reduce the number of getters and make it easier to read.

Graphics/GraphicsEngineVulkan/src/PipelineLayoutVk.cpp
```diff
@@ -92,6 +100,29 @@ void PipelineLayoutVk::Create(RenderDeviceVkImpl* pDeviceVk, RefCntAutoPtr<Pipel
...
+                    pPushConstantResDesc         = &ResDesc;
+                    m_PushConstantSignatureIndex = BindInd;
+                    m_PushConstantResourceIndex  = r;
+                    // For inline constants, ArraySize contains the number of 32-bit constants.
+                    m_PushConstantSize = ResDesc.ArraySize * sizeof(Uint32);
+                    for (SHADER_TYPE ShaderTypes = ResDesc.ShaderStages; ShaderTypes != SHADER_TYPE_UNKNOWN;)
+                    {
+                        const SHADER_TYPE ShaderType = ExtractLSB(ShaderTypes);
+                        m_PushConstantStageFlags |= ShaderTypeToVkShaderStageFlagBit(ShaderType);
+                    }
```
TheMostDiligent: There is `ShaderTypesToVkShaderStageFlags` function for this

Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
hzqst: > It is used to patch Vertex Shader inputs, **which is done when the shader is created**.

it can be a problem when ShaderVkImpl is asynchronously loaded, `SPIRVShaderResources` is not available at that time in `ShaderVkImpl::Initialize`. The ArchiverTest said that no `SPIRVShaderResources` is created when shader is created.

btw isn't shader patching done in `RemapOrVerifyShaderResources`?

Graphics/GraphicsEngineVulkan/src/PipelineLayoutVk.cpp
hzqst: Fixed in #751

Graphics/GraphicsEngineVulkan/include/PipelineLayoutVk.hpp
hzqst: Fixed in #751

Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
hzqst: we just leave it false or?

hzqst:
Seems that we have `PipelineResourceSignatureVkImpl::HasInlineConstants` failed and causing `m_PushConstantInfo` to be left empty. what do we need next?

```cpp
    bool HasInlineConstants() const
    {
        // TODO
        return false;
    }
```

TheMostDiligent:
> Seems that we have `PipelineResourceSignatureVkImpl::HasInlineConstants` failed and causing `m_PushConstantInfo` to be left empty. what do we need next?

This is a simple function that helps to avoid looping through all resources for all signatures even if there are no inline constants

Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
TheMostDiligent: There is an error somewhere. Asynchronous shaders must be created first before the pipeline is created. Querying anything from the shader while it is compiled is an error.

hzqst: Well, seems that after commenting out deferred SPIRVShaderResource creation everything works fine, also with ShaderResource initialized with `pShader->GetShaderResource()`

image

image

TheMostDiligent: I haven't checked this code yet, but based on comment, this block is certainly not needed. `InitPipelineLayout` is only called after all asynchronous shader work is complete.

TheMostDiligent: BTW, why is `ShaderResources` needed to be added to `ShaderStageInfo`?

hzqst: > BTW, why is `ShaderResources` needed to be added to `ShaderStageInfo`?

Because we are going to manipulate it later in the `PatchShaderConvertUniformBufferToPushConstant`. the ShaderVkImpl::m_pShaderResources is immutable theoretically , we are not suppose to manipulate it anymore once ShaderVkImpl initialized.

hzqst: I've removed redundant block in https://github.com/DiligentGraphics/DiligentCore/pull/750/changes/4037b436f070ae42d57ab6766b09cf28c32ffb7e

hzqst: Still can be a problem with `PSO_ARCHIVE_FLAG_STRIP_REFLECTION`. `m_pShaderResources` somehow become empty here

image

image

TheMostDiligent: Reflection is only stripped from the final SPIRV after all patching, push constants conversion including

hzqst: Still a problem with Async PSO too

image

image

hzqst: > Reflection is only stripped from the final SPIRV after all patching, push constants conversion including

nvm I skipped the validation in https://github.com/DiligentGraphics/DiligentCore/pull/750/commits/34af2e65bba6a55bdc5cbb3b7f03dac763a273ff. it works fine for stripped_reflection one now.

TheMostDiligent: I [refactored `PipelineStateVkImpl::ShaderStageInfo`](https://github.com/DiligentGraphics/DiligentCore/commit/f15f47864d2ca7a301fa14c691c431443af5ab9a), but now when I look at it I don't think `ShaderResources` are needed in `ShaderStageInfo` at all.
The problem is that right now, `PatchShaderConvertUniformBufferToPushConstant` is called separately from `RemapOrVerifyShaderResources`, but it should be part of `RemapOrVerifyShaderResources`.
`RemapOrVerifyShaderResources` should both patch bindings and push constants, if necessary.
If needed, it will created new resources and add them to `pDvpShaderResources`. If `pDvpShaderResources` is null, new resources are not needed.

hzqst: > I [refactored `PipelineStateVkImpl::ShaderStageInfo`](https://github.com/DiligentGraphics/DiligentCore/commit/f15f47864d2ca7a301fa14c691c431443af5ab9a), but now when I look at it I don't think `ShaderResources` are needed in `ShaderStageInfo` at all. The problem is that right now, `PatchShaderConvertUniformBufferToPushConstant` is called separately from `RemapOrVerifyShaderResources`, but it should be part of `RemapOrVerifyShaderResources`. `RemapOrVerifyShaderResources` should both patch bindings and push constants, if necessary. If needed, it will created new resources and add them to `pDvpShaderResources`. If `pDvpShaderResources` is null, new resources are not needed.

but  `pDvpShaderResources` is a  `DILIGENT_DEVELOPMENT-only` variable which is not available in production build.

TheMostDiligent: Yes, which is why in production build there is no even need to create new shader resources after patching SPIRV.

Graphics/GraphicsEngineVulkan/include/ShaderResourceCacheVk.hpp
```diff
@@ -245,6 +250,32 @@ class ShaderResourceCacheVk : public ShaderResourceCacheBase
                                 Uint32 CacheOffset,
                                 Uint32 DynamicBufferOffset);
 
+    // Gets the inline constant data pointer from the resource cache
+    const void* GetInlineConstantData(Uint32 DescrSetIndex, Uint32 CacheOffset) const;
+
+    // Initialize inline constant buffer in the resource cache
+    void InitializeInlineConstantBuffer(Uint32 DescrSetIndex,
```
TheMostDiligent: Can you rework this similar to [ShaderResourceCacheD3D12](https://github.com/DiligentGraphics/DiligentCore/blob/169b5e13a7e315500c1e3521d8e93e141097831a/Graphics/GraphicsEngineD3D12/include/ShaderResourceCacheD3D12.hpp#L157).
The fact that memory is allocated by the cache, but offsets are initialized by the signature make it first difficult to reason, and second, error-prone because the memory management logic is split between two places.

hzqst: A rework has been done in https://github.com/DiligentGraphics/DiligentCore/pull/750/changes/00024752c932d3047618610229c7e76f1f7e2e11

Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
```diff
@@ -1053,7 +1053,7 @@ void PipelineStateVkImpl::DvpValidateResourceLimits() const
             const PipelineResourceSignatureVkImpl::ResourceAttribs& ResAttr   = pSignature->GetResourceAttribs(r);
             const Uint32                                            DescIndex = static_cast<Uint32>(ResAttr.DescrType);
 
-            DescriptorCount[DescIndex] += ResAttr.ArraySize;
+            DescriptorCount[DescIndex] += ResDesc.GetArraySize();
```
TheMostDiligent: Is this needed? `ResAttr.ArraySize` should be the correct array size from the shader.

hzqst: > Is this needed? `ResAttr.ArraySize` should be the correcy array size from the shader.

it's needed, restoring `DescriptorCount[DescIndex] += ResDesc.GetArraySize();` back to `DescriptorCount[DescIndex] += ResAttr.ArraySize;` lead to unexpected `NumDynamicUniformBuffers`.

image

TheMostDiligent: On which test does it fail? Everything passes for me.
`ResAttr.ArraySize` should be equal to `ResDesc.GetArraySize()`

hzqst: > does it fail? Everything passes for me.

https://github.com/hzqst/DiligentCore/blob/afe0f504bc339dccd5cf9b7f3af253bbaf1c3caa/Tests/DiligentCoreAPITest/src/InlineConstantsTest.cpp#L230

```cpp
            PipelineResourceLayoutDescX ResLayoutDesc;
            ResLayoutDesc
                .AddVariable(SHADER_TYPE_VERTEX, "cbInlinePositions", PosType, SHADER_VARIABLE_FLAG_INLINE_CONSTANTS)
                .AddVariable(SHADER_TYPE_VS_PS, "cbInlineColors", ColType, SHADER_VARIABLE_FLAG_INLINE_CONSTANTS);

            PsoCI
                .AddRenderTarget(pSwapChain->GetDesc().ColorBufferFormat)
                .SetPrimitiveTopology(PRIMITIVE_TOPOLOGY_TRIANGLE_LIST)
                .AddShader(sm_Res.pVS)
                .AddShader(sm_Res.pPS)
                .SetResourceLayout(ResLayoutDesc);
            PsoCI.GraphicsPipeline.DepthStencilDesc.DepthEnable = False;

            RefCntAutoPtr<IPipelineState> pPSO;
            pDevice->CreateGraphicsPipelineState(PsoCI, &pPSO);
```

`DescriptorCount[static_cast<Uint32>(DescriptorType::UniformBufferDynamic)] = NumDynamicUniformBuffers = 36`
`Limits.maxDescriptorSetUniformBuffersDynamic = 8`

TheMostDiligent: For which resource does `ResAttr.ArraySize` differ from `ResDesc.GetArraySize()`?

hzqst: > For which resource does `ResAttr.ArraySize` differ from `ResDesc.GetArraySize()`?

image

image

TheMostDiligent: Hmm, both 24 and 12 are wrong. It should be 1.
For me, it is always 1, I never get anything else.
Maybe your build is stale?
Try rebuilding the `Diligent-GraphicsEngineVk-static` only

hzqst: wonder if it's intended, it was a copy-paste from `PipelineStateD3D11Impl.cpp`

image

https://github.com/hzqst/DiligentCore/blob/a8ad55b10c7b5da2fc5356962ca177f543b5cc7a/Graphics/GraphicsEngineD3D11/src/PipelineStateD3D11Impl.cpp#L99C31-L99C32

https://github.com/hzqst/DiligentCore/blob/a8ad55b10c7b5da2fc5356962ca177f543b5cc7a/Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp#L640

TheMostDiligent: I meant, `ResAttr.ArraySize` is always 1.

TheMostDiligent: But that's different. `ResDesc.GetArraySize()` must always be the same as `ResAttr.ArraySize`
Try adding assert to verify this.

hzqst: > But that's different. `ResDesc.GetArraySize()` must always be the same as `ResAttr.ArraySize` Try adding assert to verify this.

We probably gonna remove 
`ArraySize = Attribs.GetInlineConstantCountOrThrow(pShader->GetDesc().Name);`
from 
`PipelineStateD3D11Impl::GetDefaultResourceSignatureDesc`, `PipelineStateD3D12Impl::GetDefaultResourceSignatureDesc`, `PipelineStateVkImpl::GetDefaultResourceSignatureDesc` ?

TheMostDiligent: > We probably gonna remove

Why? It is needed to determine buffer size for shaders that have push constants.

hzqst: > > We probably gonna remove
> 
> Why? It is needed to determine buffer size for shaders that have push constants.

because the `ResAttr.ArraySize' '24` and `12` indeed came from `PipelineStateVkImpl::GetDefaultResourceSignatureDesc` -> `ArraySize = Attribs.GetInlineConstantCountOrThrow(pShader->GetDesc().Name);`

TheMostDiligent: Everything is correct here.
`ArraySize = Attribs.GetInlineConstantCountOrThrow(pShader->GetDesc().Name)` is used to set the `PipelineResourceDesc`, which should contain the number of inline constants, computed from the buffer size.
Resource attribs contain shader array size.

hzqst: > PipelineResourceDesc

well, got it.

the

```cpp
 DescriptorCount[DescIndex] += ResDesc.GetArraySize();
```
should be

```cpp
 DescriptorCount[DescIndex] += ResAttr.ArraySize;
```

to match the semantics or intention of the `DescriptorCount[]`

TheMostDiligent: But really they should be the same like I said.
I don't quite understand where you get the issue.

hzqst: Fixed in https://github.com/DiligentGraphics/DiligentCore/pull/750/commits/2e1b0a5582802ca253aeeb7f96f6c3789857ecd4

hzqst: > But really they should be the same like I said. I don't quite understand where you get the issue.

nvm, my eye slipped with `ResDesc` & `ResAttr`

TheMostDiligent: OK, that explains your confusion. Right, they are very similar.

TheMostDiligent:
Take a look at how much simpler I made the inline constant initialization in the resource cache.
Few and simpler changes are always preferable as they are easier to understand, and less code to maintain.

TheMostDiligent:
We need a test that [tests this path](https://github.com/DiligentGraphics/DiligentCore/blob/4c68bf7058c09f5b16f5ab0c6cb0ecdd8020c370/Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp#L640). It needs to use a shader with native push constants block, it should properly map it to inline constants in the resource signature. (Vulkan only)

hzqst:
> We need a test that [tests this path](https://github.com/DiligentGraphics/DiligentCore/blob/4c68bf7058c09f5b16f5ab0c6cb0ecdd8020c370/Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp#L640). It needs to use a shader with native push constants block, it should properly map it to inline constants in the resource signature. (Vulkan only)

test done in https://github.com/DiligentGraphics/DiligentCore/pull/750/commits/fe5ea92826cbcae4aa23574dd95ac85ce96bf1fe

Here is another problem:

What if we have `[[vk::push_constant]] ConstantBuffer<PushConstants_t> PushConstants;` declared both in VS and PS?

The `VS->PushConstants` and `PS->PushConstants` seems to be completely difference resources due to:
```cpp
const auto it_assigned = UniqueResources.emplace(ShaderResourceHashKey{VarDesc.ShaderStages, Attribs.Name}, Attribs);
```

Writing inline constants into `VS->PushConstants` won't affects `PS->PushConstants` and vice versa, which is confusing.

I suggust either merging them into identical resource, or reject the shaders where `VS->PushConstants` and `PS->PushConstants` are both existing.

TheMostDiligent:
> I suggust either merging them into identical resource, or reject the shaders where VS->PushConstants and PS->PushConstants are both existing.

Yes, if push constants block present in the shaders, they need to be identical. We will merge them into a single push constant block

TheMostDiligent:
Can you please do rebases instead of merges? Linearizing history after merge is very inconvenient

hzqst:
> Can you please do rebases instead of merges? Linearizing history after merge is very inconvenient

will do next time

Tests/DiligentCoreAPITest/src/InlineConstantsTest.cpp
```diff
@@ -100,6 +100,43 @@ float4 main(in PSInput PSIn) : SV_Target
...
+    R"(
+struct PSInput
+{
+    float4 Pos   : SV_POSITION;
+    float4 Color : COLOR;
+};
+
+float4 main(in PSInput PSIn) : SV_Target
+{
+    return PSIn.Color;
```
TheMostDiligent: Why PS is not using the push constants block? It should to verify that the stages are properly merged.

TheMostDiligent:
@hzqst Don't rebase on inline_constants yet - the tests will break due to incorrect memory size for SRB allocator.

Tests/DiligentCoreAPITest/src/InlineConstantsTest.cpp
hzqst: > Why PS is not using the push constants block? It should to verify that the stages are properly merged.

That's what I mentioned before: 

If we have `[[vk::push_constant]] ConstantBuffer<PushConstants_t> PushConstants;` declared both in VS and PS, they will be allocated as two resources instead of one, in `PipelineStateVkImpl::GetDefaultResourceSignatureDesc`.

Calling `pPSO->GetStaticVariableByName(SHADER_TYPE_PIXEL, "PushConstants")->SetInlineConstants` does nothing under the hood as data never get committed in `m_CommandBuffer.PushConstants(Layout.GetVkPipelineLayout(), PushConstantInfo.vkRange, pPushConstantData);`. The `PushConstantInfo.vkRange` goes with nothing more than `VK_SHADER_STAGE_VERTEX_BIT`.

image

I suggust ensure that `[[vk::push_constant]] ConstantBuffer<PushConstants_t> PushConstants;` has exact the same size in both VS and PS first, if they are both found in VS and PS.

We merge `PCInfo.vkRange.stageFlags` with `ShaderTypesToVkShaderStageFlags(SHADER_TYPE_PIXEL)` in the later pass in `PipelineLayoutVk::GetPushConstantInfo` then.

image

`PS inline constants` need to be merged into `VS inline constants` in `PipelineStateVkImpl::GetDefaultResourceSignatureDesc` as an identical resource also.

TheMostDiligent:
I am going to rebase and overwrite your branch - give me a few minutes.
You will need to submit the last change again - it is not yet working anyway.

TheMostDiligent:
OK, done

hzqst:
@TheMostDiligent we are now error out with https://github.com/DiligentGraphics/DiligentCore/pull/750/commits/9aad908ea33b87963f8c5e2eb6ae2f3b98fd0fb6

image


If we have `[[vk::push_constant]] ConstantBuffer<PushConstants_t> PushConstants;` declared both in VS and PS, they will be allocated as two resources instead of one, in `PipelineStateVkImpl::GetDefaultResourceSignatureDesc`.

Calling `pPSO->GetStaticVariableByName(SHADER_TYPE_PIXEL, "PushConstants")->SetInlineConstants` does nothing under the hood as data never get committed in `m_CommandBuffer.PushConstants(Layout.GetVkPipelineLayout(), PushConstantInfo.vkRange, pPushConstantData);`. The `PushConstantInfo.vkRange` goes with nothing more than `VK_SHADER_STAGE_VERTEX_BIT`.

image

I suggust ensure that `[[vk::push_constant]] ConstantBuffer<PushConstants_t> PushConstants;` has exact the same size in both VS and PS first, if they are both found in VS and PS.

We should probably merge `PCInfo.vkRange.stageFlags` with `ShaderTypesToVkShaderStageFlags(SHADER_TYPE_PIXEL)` later in `PipelineLayoutVk::GetPushConstantInfo`.

image

`PS inline constants` as a `PRSDesc` need to be merged into `VS inline constants` in `PipelineStateVkImpl::GetDefaultResourceSignatureDesc` also.

TheMostDiligent:
> I suggust ensure that [[vk::push_constant]] ConstantBuffer<PushConstants_t> PushConstants; has exact the same size in both VS and PS first, if they are both found in VS and PS.

Yes, this is what we need to do.
Instead of `break`, the loop should check if PCInfo is not empty already, and if so:
- Check if the name and size matches, then add the shader stage flag
- Otherwise throw an error

Can you work on this in a separate PR?
I want to finish merging this one.

TheMostDiligent:
I will also remove the test from this branch.

hzqst:
> > I suggust ensure that [[vk::push_constant]] ConstantBuffer<PushConstants_t> PushConstants; has exact the same size in both VS and PS first, if they are both found in VS and PS.
> 
> Yes, this is what we need to do. Instead of `break`, the loop should check if PCInfo is not empty already, and if so:
> 
> * Check if the name and size matches, then add the shader stage flag
> * Otherwise throw an error
> 
> Can you work on this in a separate PR? I want to finish merging this one.

base on [DiligentGraphics:inline_constants](https://github.com/DiligentGraphics/DiligentCore/tree/inline_constants) or [hzqst:inline_constants_747_fix](https://github.com/hzqst/DiligentCore/tree/inline_constants_747_fix)?

TheMostDiligent:
On [hzqst:inline_constants_747_fix](https://github.com/hzqst/DiligentCore/tree/inline_constants_747_fix), [DiligentGraphics:inline_constants](https://github.com/DiligentGraphics/DiligentCore/tree/inline_constants) don't have full implementation yet.

TheMostDiligent:
Can you please create another PR?
I am working on merging this one.

hzqst:
Done in https://github.com/DiligentGraphics/DiligentCore/pull/750/commits/b0e30b58b9032efbb98e636e03c56230a2fbd96d

Next problem: what we gonna do with `PushConstantInfo::SignatureIndex` and `PushConstantInfo::ResourceIndex`? we have two `ResourceIndex` now, if merge happens.

Graphics/GraphicsEngineVulkan/src/DeviceContextVkImpl.cpp
```diff
@@ -536,9 +577,30 @@ void DeviceContextVkImpl::CommitShaderResources(IShaderResourceBinding* pShaderR
...
+
+    // If there are no descriptor sets, we're done (SRB only contains push constants)
+    // Clear the stale flag since there are no descriptor sets to commit
     if (ResourceCache.GetNumDescriptorSets() == 0)
     {
-        // Ignore SRBs that contain no resources
+        // Clear the stale bit for this SRB since it has no descriptor sets to commit
+        // The SRB is still bound and ResourceCaches[SRBIndex] is set, so push constants can be accessed
+        const auto SRBBit = static_cast<ResourceBindInfo::SRBMaskType>(1u << SRBIndex);
+        BindInfo.StaleSRBMask &= ~SRBBit;
```
TheMostDiligent: Why are all these changed needed?
If I revert all of them, everything works fine.

TheMostDiligent: Based on comments, it looks like this is remained from your earlier versions, since now all inline constants get descriptor sets, this does not make sense:
`// Always bind the SRB, even if it has no descriptor sets (e.g., only push constants)`
Also, 
```
    PIPELINE_TYPE SRBPipelineType = pResBindingVkImpl->GetPipelineType();
    if (SRBPipelineType == PIPELINE_TYPE_INVALID)
        return;
```
Should never happen.

PR #752, opened 2026-01-08
Test VulkanPushConstantsBlock, merge stages when both VS & PS found accessing PushConstants.

hzqst:
Based on https://github.com/DiligentGraphics/DiligentCore/pull/750#issuecomment-3717781007

`PipelineLayoutVk::GetPushConstantInfo`: `PCInfo.vkRange.stageFlags` now merges `VK_SHADER_STAGE_VERTEX_BIT | VK_SHADER_STAGE_FRAGMENT_BIT` , instead of `VK_SHADER_STAGE_VERTEX_BIT `.

`PipelineStateVkImpl::GetDefaultResourceSignatureDesc`: VS "PushConstants" and PS "PushConstants" now being loaded as an identical resource signature with `SHADER_TYPE_VS_PS`, instead of `SHADER_TYPE_VERTEX` + `SHADER_TYPE_PIEXEL`.

### DIFF
--- a/Tests/DiligentCoreAPITest/src/InlineConstantsTest.cpp
+++ b/Tests/DiligentCoreAPITest/src/InlineConstantsTest.cpp
@@ -459,18 +459,15 @@ TEST_F(InlineConstants, ComputeResourceLayout)
         // Set static inline constants on PSO before creating SRB
         if (ResType == SHADER_RESOURCE_VARIABLE_TYPE_STATIC)
         {
-            IShaderResourceVariable* pVar = pPSO->GetStaticVariableByName(SHADER_TYPE_COMPUTE, "cbInlineData");
-            ASSERT_TRUE(pVar);
-            pVar->SetInlineConstants(g_ComputeData, 0, kNumComputeConstants);
+            IShaderResourceVariable* pInlineDataVar = pPSO->GetStaticVariableByName(SHADER_TYPE_COMPUTE, "cbInlineData");
+            ASSERT_NE(pInlineDataVar, nullptr);
+            pInlineDataVar->SetInlineConstants(g_ComputeData, 0, kNumComputeConstants);
         }
 
-        // Set static UAV on PSO
         {
-            IShaderResourceVariable* pVar = pPSO->GetStaticVariableByName(SHADER_TYPE_COMPUTE, "g_Output");
-            if (pVar != nullptr)
-            {
-                pVar->Set(pOutputUAV);
-            }
+            IShaderResourceVariable* pOutputVar = pPSO->GetStaticVariableByName(SHADER_TYPE_COMPUTE, "g_Output");
+            ASSERT_NE(pOutputVar, nullptr);
+            pOutputVar->Set(pOutputUAV);
         }
 
         // Create SRB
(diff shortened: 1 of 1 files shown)


TheMostDiligent:
This seems to be too complicated, I think I mislead you with how it should be implemented.
I think the only thing that needs to be done is merge inline constant variables here (i.e. add stage flag to existing variable instead of adding new):

https://github.com/DiligentGraphics/DiligentCore/blob/dae660a320830702a263fc6e96847ba50c173237/Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp#L635-L641

Then `PipelineLayoutVk::GetPushConstantInfo` will not need to be changed at all as it will use proper merged stages.

TheMostDiligent:
Here: aa7539a4c2cd60847601e2df72eb9e89d16f6fdb

Tests/DiligentCoreAPITest/src/InlineConstantsTest.cpp
```diff
@@ -772,4 +819,131 @@ TEST_F(InlineConstants, RenderStateCache)
...
+        float4{+0.5f, +0.5f, 0.f, 1.f},
+        float4{+1.0f, -0.5f, 0.f, 1.f},
+    };
+
+    float4 PushConstants_Colors[] = {
+        float4{1.f, 0.f, 0.f, 1.f},
+        float4{0.f, 1.f, 0.f, 1.f},
+        float4{0.f, 0.f, 1.f, 1.f},
+        float4{1.f, 1.f, 1.f, 1.f},
+    };
```
TheMostDiligent: There are already global constants for this

Tests/DiligentCoreAPITest/src/InlineConstantsTest.cpp
```diff
@@ -772,4 +819,131 @@ TEST_F(InlineConstants, RenderStateCache)
...
+    const float ClearColor[] = {sm_Rnd(), sm_Rnd(), sm_Rnd(), sm_Rnd()};
+    RenderDrawCommandReference(pSwapChain, ClearColor);
+
+    ITextureView* pRTVs[] = {pSwapChain->GetCurrentBackBufferRTV()};
+    pContext->SetRenderTargets(1, pRTVs, nullptr, RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
+    pContext->ClearRenderTarget(pRTVs[0], ClearColor, RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
+
+    //They should all goes with resource[0] ("PushConstants") under the hood:
+
+    IShaderResourceVariable* pVSVar = pPSO->GetStaticVariableByName(SHADER_TYPE_VERTEX, "PushConstants");
```
TheMostDiligent: Let's extend the test a bit - test all variable types by specifying
`PsoCI.PSODesc.ResourceLayout.DefaultVariableType`

TheMostDiligent (changes requested):

TheMostDiligent:
Can you please also add a test similar to `InlineConstants.ResourceLayout`, but for compute shaders - they are not covered at all now.

hzqst:
> * Requested changes

Done in https://github.com/DiligentGraphics/DiligentCore/pull/752/changes/3cfb920b3a38a35138d1f68656922a270458e310

TheMostDiligent:
Just another remined to not use merges, they mess up history a lot

PR #755, opened 2026-01-12
Fix Inline Constants Cross-Signature Commit Inconsistency for vulkan impl

hzqst:
See [SetInlineConstants_InconsistencyCommit](https://github.com/hzqst/DiligentCore/blob/inline_constants_claude/plan/SetInlineConstants_InconsistencyCommit.md)


TheMostDiligent:
We need a unit test for this. Every bug/issue/new feature should be covered.
I think it can be reproduced this way:
- Create two compatible signatures
- Use signature 1 to create PSO
- Use signature 2 to create SRB

PR #756, opened 2026-01-12
Add CrossSignatureSRB test, and corresponding fix for it.

hzqst:
See [SetInlineConstants_InconsistencyCommit](https://github.com/hzqst/DiligentCore/blob/inline_constants_claude/plan/SetInlineConstants_InconsistencyCommit.md)

### DIFF
--- a/Tests/DiligentCoreAPITest/src/InlineConstantsTest.cpp
+++ b/Tests/DiligentCoreAPITest/src/InlineConstantsTest.cpp
@@ -32,7 +32,7 @@
 #include "RenderStateCache.hpp"
 #include "GraphicsTypesX.hpp"
 #include "FastRand.hpp"
-
+#include "MapHelper.hpp"
 
 
 namespace Diligent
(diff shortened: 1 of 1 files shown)


Tests/DiligentCoreAPITest/src/InlineConstantsTest.cpp
```diff
@@ -839,6 +839,179 @@ TEST_F(InlineConstants, TwoResourceSignatures)
...
+// 5. BUG (fixed): UpdateInlineConstantBuffers() was updating Signature A's buffers
+//    instead of the buffers from SRB cache
+//
+// The fix ensures that UpdateInlineConstantBuffers() updates the buffer from
+// SRB cache - the same buffer that was bound during BindResources().
+//
+// In Vulkan, only the first inline constant uses push constants (always works).
+// The second inline constant uses buffer-emulated path where this bug manifested.
+// Therefore, this test uses TWO inline constant buffers to ensure we test both paths.
+TEST_F(InlineConstants, CrossSignatureSRB)
```
TheMostDiligent, line 862: I think this test may be simpler. Just one signature and one variable should be enough. We already test different variable combinations.

TheMostDiligent:
Your test did not actually reproduce the problem because there was only one dynamic buffer mapped in the frame, and it got zero offset. So it worked even when buffer from another signature was used because it also had zero offset.

PR #757, opened 2026-01-13
Enable SetInlineConstants for WebGPU

hzqst:


Graphics/GraphicsEngineWebGPU/src/DeviceContextWebGPUImpl.cpp
```diff
@@ -379,12 +396,13 @@ void DeviceContextWebGPUImpl::Draw(const DrawAttribs& Attribs)
 {
     TDeviceContextBase::Draw(Attribs, 0);
 
+    if (Attribs.NumVertices == 0 || Attribs.NumInstances == 0)
```
TheMostDiligent: This is incorrect. `DvpValidateCommittedShaderResources()` is now only called for empty draw calls, which is not the purpose.

Graphics/GraphicsEngineWebGPU/src/DeviceContextWebGPUImpl.cpp
```diff
@@ -394,12 +412,13 @@ void DeviceContextWebGPUImpl::MultiDraw(const MultiDrawAttribs& Attribs)
 {
     TDeviceContextBase::MultiDraw(Attribs, 0);
 
+    if (Attribs.NumInstances == 0)
```
TheMostDiligent: Same here

Graphics/GraphicsEngineWebGPU/src/DeviceContextWebGPUImpl.cpp
```diff
@@ -414,12 +433,13 @@ void DeviceContextWebGPUImpl::DrawIndexed(const DrawIndexedAttribs& Attribs)
 {
     TDeviceContextBase::DrawIndexed(Attribs, 0);
 
+    if (Attribs.NumIndices == 0 || Attribs.NumInstances == 0)
```
TheMostDiligent: Same here

Graphics/GraphicsEngineWebGPU/src/DeviceContextWebGPUImpl.cpp
```diff
@@ -429,12 +449,13 @@ void DeviceContextWebGPUImpl::MultiDrawIndexed(const MultiDrawIndexedAttribs& At
 {
     TDeviceContextBase::MultiDrawIndexed(Attribs, 0);
 
+    if (Attribs.NumInstances == 0)
```
TheMostDiligent: Same here

Graphics/GraphicsEngineWebGPU/src/DeviceContextWebGPUImpl.cpp
```diff
@@ -2098,7 +2120,12 @@ WGPURenderPassEncoder DeviceContextWebGPUImpl::PrepareForDraw(DRAW_FLAGS Flags)
     }
 
     if (CommittedShaderResources::SRBMaskType CommitSRBMask = m_BindInfo.GetCommitMask(Flags & DRAW_FLAG_DYNAMIC_RESOURCE_BUFFERS_INTACT, Flags & DRAW_FLAG_INLINE_CONSTANTS_INTACT))
-        CommitBindGroups(wgpuRenderCmdEncoder, CommitSRBMask);
+        CommitBindGroups(wgpuRenderCmdEncoder, CommitSRBMask, (Flags & DRAW_FLAG_INLINE_CONSTANTS_INTACT) != 0);
+
+#ifdef DILIGENT_DEVELOPMENT
+    // Must be called after CommitBindGroups as inline constant buffers must be updated first
+    DvpValidateCommittedShaderResources();
```
TheMostDiligent: What is the issue with inline constants?
Validation must be performed **before** committing bind groups because there might be problems that may cause it to crash and there will be no error messages.

TheMostDiligent (changes requested):

Graphics/GraphicsEngineWebGPU/src/DeviceContextWebGPUImpl.cpp
hzqst: > What is the issue with inline constants? Validation must be performed **before** committing bind groups because there might be problems that may cause it to crash and there will be no error messages.

I was debugging an error in `TEST_F(InlineConstants, ResourceLayout1)`, when `pPosVar == nullptr` (STATIC type), an error occurs: "Dynamic allocation of dynamic buffer 'Implicit signature of PSO 'Inline constants test' - cbInlinePositions' in frame 2 is out-of-date."

   - Vulkan: `DvpValidateCommittedShaderResources()` is called **AFTER** `CommitInlineConstants()` and `CommitDescriptorSets()` inside `PrepareForDraw()`

   - WebGPU: `DvpValidateCommittedShaderResources()` **was** called **BEFORE** `PrepareForDraw()`, but inline constant buffers are updated INSIDE `PrepareForDraw()`

Call `DvpValidateCommittedShaderResources` after `CommitBindGroups` just to align with vulkan impl. This also fixed the issue with "frame 2 is out-of-date" .

image
image

Graphics/GraphicsEngineWebGPU/src/PipelineStateWebGPUImpl.cpp
```diff
@@ -778,7 +778,29 @@ PipelineResourceSignatureDescWrapper PipelineStateWebGPUImpl::GetDefaultResource
...
+                    Uint32 ArraySize = Attribs.ArraySize;
+                    if (Flags & PIPELINE_RESOURCE_FLAG_INLINE_CONSTANTS)
+                    {
+                        VERIFY(Flags == PIPELINE_RESOURCE_FLAG_INLINE_CONSTANTS, "INLINE_CONSTANTS flag cannot be combined with other flags.");
+                        // For inline constants, ArraySize must be the number of 32-bit constants
+                        if (Attribs.BufferStaticSize == 0)
+                        {
+                            LOG_ERROR_AND_THROW("Unable to determine inline constant count for uniform buffer '",
+                                                Attribs.Name, "'. Please enable ShaderCreateInfo.LoadConstantBufferReflection.");
+                        }
```
TheMostDiligent: This should not be possible - BufferStaticSize is always initialized

PR #768, opened 2026-02-22
SpecializationConstant impl stage 1 - API and feature plumbing

hzqst:
Plan: [SpecializationConstants.md](https://github.com/hzqst/DiligentCore/blob/master-claude/plan/SpecializationConstants.md)

### Objective
Introduce a backend-agnostic API surface for specialization constants and add feature gating so unsupported backends fail fast.

### Work items
1. **Public API type additions**
   - Add `SpecializationConstant` in `Graphics/GraphicsEngine/interface/PipelineState.h`:
     - `const char* Name`
     - `SHADER_TYPE ShaderStages`
     - `Uint32 Size`
     - `const void* pData`
   - Extend `PipelineStateCreateInfo` with:
     - `const SpecializationConstant* pSpecializationConstants`
     - `Uint32 NumSpecializationConstants`
   - Add `SpecializationConstants` to `DeviceFeatures` in `Graphics/GraphicsEngine/interface/GraphicsTypes.h`.
   - Add `SpecializationConstants` to `ENUMERATE_DEVICE_FEATURES` and constructor initialization paths.
   - Update all `ASSERT_SIZEOF(DeviceFeatures, 47, ...)` call sites to the new size.
   - Wire feature enable/selection logic in `Graphics/GraphicsEngine/src/RenderDeviceBase.cpp` (`EnableDeviceFeatures`).
   
2. **API validation and behavior contract**
   - Add common validation in pipeline create-info validation path (`PipelineStateBase` validation):
     - pointer/count consistency (`Num == 0` <=> pointer may be null, `Num > 0` requires pointer non-null)
     - every entry must have valid `Name`, `ShaderStages`, `Size`, `pData`
     - reject duplicate `(Name, ShaderStages overlap)` in one PSO create info
     - reject usage when `DeviceFeatures.SpecializationConstants == DEVICE_FEATURE_STATE_DISABLED`
   - Keep behavior deterministic across all PSO types (graphics/compute/ray tracing/tile where applicable).

3. **Backend feature states**
   - Vulkan: set `Features.SpecializationConstants = ENABLED`.
   - WebGPU: set `Features.SpecializationConstants = ENABLED`.
   - D3D11 / D3D12 / OpenGL / Metal: set `DISABLED`.
   - Update backend feature-population files (not exhaustive):
     - `Graphics/GraphicsEngineVulkan/src/VulkanTypeConversions.cpp`
     - `Graphics/GraphicsEngineWebGPU/src/EngineFactoryWebGPU.cpp`
     - `Graphics/GraphicsEngineD3D11/src/EngineFactoryD3D11.cpp`
     - `Graphics/GraphicsEngineD3D12/src/EngineFactoryD3D12.cpp`
     - `Graphics/GraphicsEngineOpenGL/src/RenderDeviceGLImpl.cpp`
     - `Graphics/GraphicsEngineVulkan/src/EngineFactoryVk.cpp` (size/assert sync)

4. **Create-info lifetime, copy, and serialization**
   - Extend `PipelineStateCreateInfoX` wrapper in `Graphics/GraphicsEngine/interface/GraphicsTypesX.hpp` to deep-copy specialization constants (names + raw bytes) for async PSO creation safety.
   - Extend PSO serialization/deserialization for `PipelineStateCreateInfo`:
     - `Graphics/GraphicsEngine/src/PSOSerializer.cpp`
     - include per-entry `{Name, ShaderStages, Size, DataBytes}`.
   - Update serialization size asserts and tests.

5. **Hash combiner update for render state cache**
   - Update `HashCombiner<HasherType, PipelineStateCreateInfo>` in `Common/interface/HashUtils.hpp` to include `NumSpecializationConstants` and each entry's `{Name, ShaderStages, Size, DataBytes}` in the hash.
   - Without this, the render state cache (`RenderStateCacheImpl`) cannot distinguish PSOs that differ only by specialization constants, since the PSO lookup key (`XXH128Hash`) would collide.

6. **Tests for stage 1**
   - Update `Tests/DiligentCoreTest/src/GraphicsEngine/PSOSerializerTest.cpp` for round-trip with non-empty specialization constants.
   - Add API validation tests in `Tests/DiligentCoreAPITest/src/` for each error path about SpecializationConstants:
     - null pointer with non-zero count
     - feature disabled rejection
     - null / empty `Name`, unknown `ShaderStages`, zero `Size`, null `pData`
     - duplicate `Name` with overlapping `ShaderStages`
   - Add `GraphicsTypesXTest` coverage for `AddSpecializationConstant`, `ClearSpecializationConstants`, copy construction, and move construction with specialization constants attached.
   - Add render state cache tests (`Tests/DiligentCoreAPITest/src/RenderStateCacheTest.cpp`) that create PSOs with specialization constants:
     - Verify that a PSO with spec constants can be cached and retrieved.
     - Verify that PSOs with different spec constant values produce distinct cache entries (not hash collisions).
     - Follow the existing `InlineConstants::RenderStateCache` pattern (3-pass: empty cache → loaded → reloaded).


Tests/DiligentCoreAPITest/src/SpecializationConstantsTest.cpp
```diff
@@ -0,0 +1,283 @@
+/*
```
TheMostDiligent, line 1: There are dedicated PSO creation failure tests for this purpose:

https://github.com/DiligentGraphics/DiligentCore/blob/master/Tests/DiligentCoreAPITest/src/ObjectCreationFailure/PSOCreationFailureTest.cpp

Tests/DiligentCoreAPITest/src/RenderStateCacheTest.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2019-2025 Diligent Graphics LLC
+ *  Copyright 2019-2026 Diligent Graphics LLC
```
TheMostDiligent, line 2: Render state cache tests don't make a lot of sense at this point - let's add them when the constants are actually implemented, so that we can test that constants are properly set in the pipeline.

Also, with the cache, with current implementation the pipelines will be cached for every combination of spec constants, which is a bit suboptimal. Ideally, we only cache pipeline once and set initialization constants when it is extracted.

This does not need to be done immediately though. Let's make it work first and see how to optimize later.

TheMostDiligent (changes requested):

Tests/DiligentCoreAPITest/src/SpecializationConstantsTest.cpp
hzqst: ported to `PSOCreationFailureTest.cpp` in f974a54ab3ee0212a1624791eb5fd2aa46db8b2d

Tests/DiligentCoreAPITest/src/RenderStateCacheTest.cpp
hzqst: Do I need to remove them all from current branch or just leave them there?

TheMostDiligent:
I merged failure tests, this part is complete, so the PR can be closed.
For the next step, we can do `SPIRVShaderResources` updates and tests.

PR #769, opened 2026-02-25
Add specialization constants support to SPIRVShaderResources.

hzqst:
Following workitem from [SpecializationConstants.md](https://github.com/hzqst/DiligentCore/blob/master-claude/plan/SpecializationConstants.md) has been implemented:

1. **SPIR-V reflection for specialization constants**
   - Extend `SPIRVShaderResources` to expose specialization constants metadata:
     - constant name (`OpName`)
     - `SpecId`
     - scalar type / byte size metadata for validation
   - Candidate files:
     - `Graphics/ShaderTools/include/SPIRVShaderResources.hpp`
     - `Graphics/ShaderTools/src/SPIRVShaderResources.cpp`
   - Use SPIRV-Cross specialization constant reflection (`Compiler.get_specialization_constants` + name lookup).
   - Need to be tested in `Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp`

### DIFF
--- a/Graphics/ShaderTools/src/SPIRVShaderResources.cpp
+++ b/Graphics/ShaderTools/src/SPIRVShaderResources.cpp
@@ -614,19 +614,12 @@ SPIRVShaderResources::SPIRVShaderResources(IMemoryAllocator&     Allocator,
     static_assert(Uint32{SPIRVShaderResourceAttribs::ResourceType::NumResourceTypes} == 13, "Please set the new resource type counter here");
 
     // Specialization constants reflection
-    struct SpecConstInfo
-    {
-        std::string            Name;
-        uint32_t               SpecId    = 0;
-        uint32_t               Size      = 0;
-        SHADER_CODE_BASIC_TYPE BasicType = SHADER_CODE_BASIC_TYPE_UNKNOWN;
-    };
-    std::vector<SpecConstInfo> SpecConstants;
-    Uint32                     NumSpecConstants = 0;
-
+    diligent_spirv_cross::SmallVector<SPIRVSpecializationConstantAttribs> SpecConstants;
+    Uint32                                                                NumSpecConstants = 0;
     {
         diligent_spirv_cross::SmallVector<diligent_spirv_cross::SpecializationConstant> spec_consts =
             Compiler.get_specialization_constants();
+        SpecConstants.reserve(spec_consts.size());
         for (const diligent_spirv_cross::SpecializationConstant& sc : spec_consts)
         {
             const diligent_spirv_cross::SPIRConstant& Constant = Compiler.get_constant(sc.id);
@@ -641,22 +634,23 @@ SPIRVShaderResources::SPIRVShaderResources(IMemoryAllocator&     Allocator,
                 continue;
             }
 
-            SpecConstInfo Info;
-            Info.Name   = Compiler.get_name(sc.id);
-            Info.SpecId = sc.constant_id;
-            // OpTypeBool has width==1 in SPIRV-Cross; use 4 bytes (VkBool32) for bool specialization constants
-            Info.Size      = Type.basetype == diligent_spirv_cross::SPIRType::Boolean ? 4 : Type.width / 8;
-            Info.BasicType = SpirvBaseTypeToShaderCodeBasicType(Type.basetype);
-
-            if (Info.Name.empty())
+            const std::string& Name = Compiler.get_name(sc.id);
+            if (Name.empty())
             {
                 LOG_WARNING_MESSAGE("Specialization constant with SpecId=", sc.constant_id,
                                     " in shader '", CI.Name, "' has no name (OpName) and will be skipped.");
                 continue;
             }
+            ResourceNamesPoolSize += Name.length() + 1;
 
-            ResourceNamesPoolSize += Info.Name.length() + 1;
-            SpecConstants.emplace_back(std::move(Info));
+            SPIRVSpecializationConstantAttribs Info{
+                Name.c_str(),
+                sc.constant_id,
+                // OpTypeBool has width==1 in SPIRV-Cross; use 4 bytes (VkBool32) for bool specialization constants
+                Type.basetype == diligent_spirv_cross::SPIRType::Boolean ? 4 : Type.width / 8,
+                SpirvBaseTypeToShaderCodeBasicType(Type.basetype),
+            };
+            SpecConstants.push_back(Info);
         }
         NumSpecConstants = static_cast<Uint32>(SpecConstants.size());
     }
@@ -893,7 +887,7 @@ SPIRVShaderResources::SPIRVShaderResources(IMemoryAllocator&     Allocator,
     if (!SpecConstants.empty())
     {
         Uint32 CurrSpecConst = 0;
-        for (const SpecConstInfo& SC : SpecConstants)
+        for (const SPIRVSpecializationConstantAttribs& SC : SpecConstants)
         {
             new (&GetSpecConstant(CurrSpecConst++)) SPIRVSpecializationConstantAttribs //
                 {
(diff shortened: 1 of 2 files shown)


Tests/DiligentCoreAPITest/src/SpecializationConstantsTest.cpp
```diff
@@ -0,0 +1,31 @@
+/*
+ *  Copyright 2019-2026 Diligent Graphics LLC
+ *
```
TheMostDiligent: This file seems to remain from previous PR

hzqst: removed.

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
```diff
@@ -589,4 +589,71 @@ TEST_F(SPIRVShaderResourcesTest, MixedResources_DXC)
...
+        {"g_UintParam", 2, 4, SHADER_CODE_BASIC_TYPE_UINT},
+        {"g_FloatParam", 3, 4, SHADER_CODE_BASIC_TYPE_FLOAT},
+    };
+
+    const SPIRVShaderResources& ConstResources = Resources;
+    EXPECT_EQ(ConstResources.GetNumSpecConstants(), static_cast<Uint32>(RefSpecConstants.size()));
+
+    // Build a map from name to reference for order-independent matching
+    std::unordered_map<std::string, const SPIRVSpecConstRefAttribs*> RefMap;
+    for (const auto& Ref : RefSpecConstants)
```
TheMostDiligent: We use auto only where it is reasonable (like when getting a map iterator). Here, direct type can be used, which will make code easier to read

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
```diff
@@ -589,4 +589,71 @@ TEST_F(SPIRVShaderResourcesTest, MixedResources_DXC)
...
+        RefMap[Ref.Name] = &Ref;
+
+    for (Uint32 i = 0; i < ConstResources.GetNumSpecConstants(); ++i)
+    {
+        const auto& SC = ConstResources.GetSpecConstant(i);
```
TheMostDiligent: Here `auto` is not need too

Graphics/ShaderTools/src/SPIRVShaderResources.cpp
```diff
@@ -613,9 +613,57 @@ SPIRVShaderResources::SPIRVShaderResources(IMemoryAllocator&     Allocator,
     ResCounters.NumPushConstants = static_cast<Uint32>(resources.push_constant_buffers.size());
     static_assert(Uint32{SPIRVShaderResourceAttribs::ResourceType::NumResourceTypes} == 13, "Please set the new resource type counter here");
 
+    // Specialization constants reflection
+    struct SpecConstInfo
+    {
+        std::string            Name;
+        uint32_t               SpecId;
+        uint32_t               Size;
+        SHADER_CODE_BASIC_TYPE BasicType;
```
TheMostDiligent: Let's make this const or add default intializers.

hzqst: Fixed in ae5362e11bd0ab111607ed446ab2a629cb7ddb82.

Tests/DiligentCoreTest/src/ShaderTools/SPIRVShaderResourcesTest.cpp
hzqst: Fixed in ae5362e11bd0ab111607ed446ab2a629cb7ddb82.

hzqst:
Btw, do we need sth. like:

```cpp
    /// When enabled, the engine will extract specialization constant names
    /// and types from the SPIR-V bytecode during shader creation.
    /// This information is required when using SpecializationConstant
    /// entries (matched by name) at pipeline state creation time.

    /// \note Loading specialization constant reflection introduces some overhead,
    ///       and should be disabled when it is not needed.
    bool LoadSpecializationConstants DEFAULT_INITIALIZER(false);
```

in `ShaderCreateInfo` ?

Graphics/ShaderTools/include/SPIRVShaderResources.hpp
```diff
@@ -203,6 +227,7 @@ class SPIRVShaderResources
         const char* CombinedSamplerSuffix       = nullptr;
         bool        LoadShaderStageInputs       = false;
         bool        LoadUniformBufferReflection = false;
+        bool        LoadSpecializationConstants = false;
```
TheMostDiligent: This parameter should not be needed. `LoadUniformBufferReflection` was added because parsing internal uniform buffer structure adds some overhead.

It looks like `Compiler.get_specialization_constants()` is not free either, but looks like it will only iterate over SPIR constants, which should be acceptable.

Otherwise we will have to add the same parameter to ShaderCreateInfo, which will be inconvenient.

Graphics/ShaderTools/src/SPIRVShaderResources.cpp
```diff
@@ -613,9 +613,57 @@ SPIRVShaderResources::SPIRVShaderResources(IMemoryAllocator&     Allocator,
...
+        uint32_t               Size      = 0;
+        SHADER_CODE_BASIC_TYPE BasicType = SHADER_CODE_BASIC_TYPE_UNKNOWN;
+    };
+    std::vector<SpecConstInfo> SpecConstants;
+    Uint32                     NumSpecConstants = 0;
+
+    if (CI.LoadSpecializationConstants)
+    {
+        auto spec_consts = Compiler.get_specialization_constants();
+        for (const auto& sc : spec_consts)
```
TheMostDiligent: Please use the type here instead of auto similar to other loops.

Graphics/ShaderTools/src/SPIRVShaderResources.cpp
```diff
@@ -613,9 +613,57 @@ SPIRVShaderResources::SPIRVShaderResources(IMemoryAllocator&     Allocator,
...
+        {
+            const diligent_spirv_cross::SPIRConstant& Constant = Compiler.get_constant(sc.id);
+            const diligent_spirv_cross::SPIRType&     Type     = Compiler.get_type(Constant.constant_type);
+
+            // Only support scalar specialization constants
+            if (Type.vecsize != 1 || Type.columns != 1)
```
TheMostDiligent, line 629: BTW, why can't we handle non-scalar constants? How are they different if we simply set raw data?

TheMostDiligent (changes requested):

Graphics/ShaderTools/src/SPIRVShaderResources.cpp
hzqst: > BTW, why can't we handle non-scalar constants? How are they different if we simply set raw data?

Because: 

image

https://docs.vulkan.org/glsl/latest/chapters/spirvmappings.html#_specialization_constants

Graphics/ShaderTools/src/SPIRVShaderResources.cpp
hzqst: Fixed in dfa48b94af072fde2b77348ab0fec7c2b084dae0

Graphics/ShaderTools/include/SPIRVShaderResources.hpp
hzqst: Fixed in d4932dde954fdc564e77242ece99a5237f610681 and 26370fc2e36dc450516218f09fa2b403373a2e9e

Issue #319, opened 2021-12-07
SparseResourceTest.SparseResidentAliasedBuffer sporadically fails in D3D12


TheMostDiligent:
SparseResourceTest.SparseResidentAliasedBuffer test sometimes fails in D3D12 when [running on CI](https://ci.appveyor.com/project/DiligentGraphics/diligentcore/builds/41794048/job/lwxsc6unrygts8ig) as well as when running locally.

PR #770, opened 2026-02-26
Enable full vulkan support of specialization constants

hzqst:
Following workitem from [SpecializationConstants.md](https://github.com/hzqst/DiligentCore/blob/master-claude/plan/SpecializationConstants.md) has been implemented:

2. **Name -> SpecId matching policy**
   - Match `SpecializationConstant::Name` to reflected specialization constant names from SPIR-V.
   - Scope matching by shader module and stage mask (`ShaderStages`).
   - Validate uniqueness and type/size compatibility.
   - Define clear error messages for:
     - name not found in target shader module
     - duplicate bindings
     - size/type mismatch

3. **Populate Vulkan structs during PSO creation**
   - In Vulkan pipeline state creation flow, build per-shader-module specialization data:
     - `std::vector<VkSpecializationMapEntry>`
     - contiguous data blob (`std::vector<Uint8>`)
     - Populate `VkSpecializationInfo`
   - Attach `StageCI.pSpecializationInfo` for each `VkPipelineShaderStageCreateInfo`.
   - Ensure memory lifetime is valid until Vulkan `vkCreate*Pipelines` call returns.
   - Primary integration file:
     - `Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp`

4. **Pipeline coverage**
   - Cover all Vulkan PSO paths that use shader stage create infos (graphics/compute/ray tracing as applicable).

5. **Tests for stage 2**
   - Add Vulkan API tests with known specialization constants:
     - graphics path and compute path minimum
     - positive test: output changes when constants change
     - negative test: unknown name / wrong size
   - Reuse existing test infrastructure patterns from inline-constants and Vulkan-specific test suites.

### DIFF
--- a/Tests/DiligentCoreAPITest/src/ObjectCreationFailure/PSOCreationFailureTest.cpp
+++ b/Tests/DiligentCoreAPITest/src/ObjectCreationFailure/PSOCreationFailureTest.cpp
@@ -1759,4 +1759,128 @@ TEST_F(PSOCreationFailureTest, SpecConst_ErrorAtSecondElement)
 }
 
 
+// ---------------------------------------------------------------------------
+// Vulkan-specific tests for BuildSpecializationData (Name->SpecId matching).
+// These require GLSL shaders with layout(constant_id) declarations so that
+// the reflected specialization constants are available for matching.
+// ---------------------------------------------------------------------------
+
+// Providing a user constant whose name does not match any reflected constant
+// in the shader should succeed silently (the constant is simply skipped).
+TEST_F(PSOCreationFailureTest, SpecConst_UnmatchedNameIsSkipped)
+{
+    auto* const pEnv       = GPUTestingEnvironment::GetInstance();
+    auto* const pDevice    = pEnv->GetDevice();
+    const auto& DeviceInfo = pDevice->GetDeviceInfo();
+
+    if (!DeviceInfo.IsVulkanDevice())
+        GTEST_SKIP() << "Name->SpecId matching is currently Vulkan-only";
+    if (DeviceInfo.Features.SpecializationConstants != DEVICE_FEATURE_STATE_ENABLED)
+        GTEST_SKIP() << "Specialization constants are not supported by this device";
+
+    // GLSL vertex shader with one specialization constant named "sc_Scale".
+    static constexpr char VSSource_GLSL[] = R"(
+        #version 450
+        layout(constant_id = 0) const float sc_Scale = 1.0;
+        #ifndef GL_ES
+        out gl_PerVertex { vec4 gl_Position; };
+        #endif
+        void main()
+        {
+            gl_Position = vec4(0.0, 0.0, 0.0, sc_Scale);
+        }
+    )";
+
+    static constexpr char PSSource_GLSL[] = R"(
+        #version 450
+        layout(location = 0) out vec4 out_Color;
+        void main()
+        {
+            out_Color = vec4(0.0, 0.0, 0.0, 1.0);
+        }
+    )";
+
+    ShaderCreateInfo ShaderCI;
+    ShaderCI.SourceLanguage = SHADER_SOURCE_LANGUAGE_GLSL_VERBATIM;
+
+    RefCntAutoPtr<IShader> pVS;
+    {
+        ShaderCI.Desc   = {"SpecConst UnmatchedName VS", SHADER_TYPE_VERTEX, true};
+        ShaderCI.Source = VSSource_GLSL;
+        pDevice->CreateShader(ShaderCI, &pVS);
+        ASSERT_TRUE(pVS);
+    }
+
+    RefCntAutoPtr<IShader> pPS;
+    {
+        ShaderCI.Desc   = {"SpecConst UnmatchedName PS", SHADER_TYPE_PIXEL, true};
+        ShaderCI.Source = PSSource_GLSL;
+        pDevice->CreateShader(ShaderCI, &pPS);
+        ASSERT_TRUE(pPS);
+    }
+
+    const float            Data         = 2.0f;
+    SpecializationConstant SpecConsts[] = {
+        // This name does not exist in the shader; it should be silently skipped.
+        {"sc_NonExistent", SHADER_TYPE_VERTEX, sizeof(Data), &Data},
+    };
+
+    auto PsoCI{GetGraphicsPSOCreateInfo("SpecConst UnmatchedNameIsSkipped")};
+    PsoCI.pVS                        = pVS;
+    PsoCI.pPS                        = pPS;
+    PsoCI.NumSpecializationConstants = _countof(SpecConsts);
+    PsoCI.pSpecializationConstants   = SpecConsts;
+
+    RefCntAutoPtr<IPipelineState> pPSO;
+    pDevice->CreateGraphicsPipelineState(PsoCI, &pPSO);
+    EXPECT_NE(pPSO, nullptr) << "Unmatched specialization constant name should be silently skipped";
+}
+
+TEST_F(PSOCreationFailureTest, SpecConst_InsufficientData)
+{
+    auto* const pEnv       = GPUTestingEnvironment::GetInstance();
+    auto* const pDevice    = pEnv->GetDevice();
+    const auto& DeviceInfo = pDevice->GetDeviceInfo();
+
+    if (!DeviceInfo.IsVulkanDevice())
+        GTEST_SKIP() << "Name->SpecId matching is currently Vulkan-only";
+    if (DeviceInfo.Features.SpecializationConstants != DEVICE_FEATURE_STATE_ENABLED)
+        GTEST_SKIP() << "Specialization constants are not supported by this device";
+
+    // GLSL compute shader with a float specialization constant (4 bytes).
+    // The spec constant must be used so the compiler does not optimize it away.
+    static constexpr char CSSource_GLSL[] = R"(
+        #version 450
+        layout(constant_id = 0) const float sc_Value = 1.0;
+        layout(rgba8, binding = 0) writeonly uniform image2D g_DummyUAV;
+        layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
+        void main()
+        {
+            imageStore(g_DummyUAV, ivec2(0), vec4(sc_Value));
+        }
+    )";
+
+    ShaderCreateInfo ShaderCI;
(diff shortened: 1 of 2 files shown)


hzqst:
@TheMostDiligent why we have unrelated error with D3D12 tests even when I did nothing to D3D12 codepath?

```
[ RUN      ] SparseResourceTest.SparseResidentAliasedBuffer
D:\a\DiligentCore\DiligentCore\Tests\GPUTestFramework\src\TestingSwapChainBase.cpp(125): error: Failed
Image rendered by the test is not identical to the reference image

[  FAILED  ] SparseResourceTest.SparseResidentAliasedBuffer (345 ms)
```

Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
```diff
@@ -95,6 +95,148 @@ void InitPipelineShaderStages(const VulkanUtilities::LogicalDevice&
...
+// Parameters:
+//   ShaderStages                - shader stages extracted from the PSO create info
+//   NumSpecializationConstants  - number of user-provided specialization constants
+//   pSpecializationConstants    - user-provided specialization constant array
+//   PSODesc                     - pipeline state description (for error messages)
+//   vkStages [in/out]           - VkPipelineShaderStageCreateInfo array to patch
+//   SpecDataPerStage [out]      - per-stage specialization data (must outlive vkCreate*Pipelines)
+//
+// Throws on validation failure (name not found, size mismatch, duplicate SpecId binding).
+void BuildSpecializationData(const PipelineStateVkImpl::TShaderStages&     ShaderStages,
```
TheMostDiligent, line 123: The logic in this function should be reversed.
At the moment, user-provided specialization constants from create info are matched to the initialization constants from shader, but it should be the opposite: for each SPIRV spec constant, a user-provided spec constant should be found.

For instance right now there will be an error if the user sets PIXEL shader stage flag because there are no spec constants in the pixel shader stage, however I don't think this needs to be an error.

The user provides the list of constants that apply to all shader stages. It should not be an error if the stage does not have constants.

I can see a scenario where an array of initialization constants may be used to initialize a set of pipelines - each pipeline will use these constants that it needs ignoring others without a failure.

Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
```diff
@@ -983,7 +1125,12 @@ void PipelineStateVkImpl::InitializePipeline(const GraphicsPipelineStateCreateIn
     std::vector<VkPipelineShaderStageCreateInfo>      vkShaderStages;
     std::vector<VulkanUtilities::ShaderModuleWrapper> ShaderModules;
 
-    InitInternalObjects(CreateInfo, vkShaderStages, ShaderModules);
+    const TShaderStages ShaderStages = InitInternalObjects(CreateInfo, vkShaderStages, ShaderModules);
+
+    // Build per-stage specialization data and patch vkShaderStages.
+    // SpecDataPerStage must outlive the vkCreate*Pipelines call below.
+    std::vector<ShaderStageSpecializationData> SpecDataPerStage;
+    BuildSpecializationData(ShaderStages, CreateInfo.NumSpecializationConstants, CreateInfo.pSpecializationConstants, m_Desc, vkShaderStages, SpecDataPerStage);
```
TheMostDiligent: I suggest moving this into InitInternalObjects to avoid repeating this logic 3 times

Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
```diff
@@ -95,6 +95,148 @@ void InitPipelineShaderStages(const VulkanUtilities::LogicalDevice&
...
+                {
+                    LOG_ERROR_AND_THROW("Description of ", GetPipelineTypeString(PSODesc.PipelineType),
+                                        " PSO '", (PSODesc.Name != nullptr ? PSODesc.Name : ""),
+                                        "' is invalid: specialization constant '", UserConst.Name,
+                                        "' was not found in ", GetShaderTypeLiteralName(Stage.Type),
+                                        " shader '", pShader->GetDesc().Name, "'.");
+                }
+
+                // Validate size compatibility.
+                if (UserConst.Size != pReflected->Size)
```
TheMostDiligent: I don't see a problem if the data is larger than what the shader expects.

Tests/DiligentCoreAPITest/src/SpecializationConstantsTest.cpp
```diff
@@ -0,0 +1,365 @@
...
+// Spec-const shader: same gradient as FillTextureCS, but channel multipliers
+// come from specialization constants.
+// Reference output: vec4(vec2(xy % 256) / 256.0, 0.0, 1.0)
+// Base color has non-zero B so that sc_MulB is not optimized away.
+// To match: sc_MulR=1.0, sc_MulG=1.0, sc_MulB=0.0
+static constexpr char g_SpecConstComputeCS_GLSL[] = R"(
+    #version 450
+    layout(constant_id = 0) const float sc_MulR = 0.0;
+    layout(constant_id = 1) const float sc_MulG = 0.0;
+    layout(constant_id = 2) const float sc_MulB = 0.0;
```
TheMostDiligent: I suggest setting sc_MulB to 1.0, or setting all them to -1.0, because for sc_MulB default value is already correct.

Tests/DiligentCoreAPITest/src/SpecializationConstantsTest.cpp
```diff
@@ -0,0 +1,365 @@
...
+    layout(constant_id = 1) const float sc_Col0_G = 0.0;
+    layout(constant_id = 2) const float sc_Col0_B = 0.0;
+
+    layout(constant_id = 3) const float sc_Col1_R = 0.0;
+    layout(constant_id = 4) const float sc_Col1_G = 1.0;
+    layout(constant_id = 5) const float sc_Col1_B = 0.0;
+
+    layout(constant_id = 6) const float sc_Col2_R = 0.0;
+    layout(constant_id = 7) const float sc_Col2_G = 0.0;
+    layout(constant_id = 8) const float sc_Col2_B = 1.0;
```
TheMostDiligent: Set default values to incorrect values to make sure they are properly set.

Tests/DiligentCoreAPITest/src/SpecializationConstantsTest.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+        gl_Position = Pos[gl_VertexIndex];
+        out_Color   = Col[gl_VertexIndex % 3];
+    }
+)";
+
+// Fragment shader: pass-through interpolated color.
+static constexpr char g_SpecConstGraphicsPS_GLSL[] = R"(
+    #version 450
+
```
TheMostDiligent, line 137: Add some spec constants here. I suggest some constant from the vertex shader + some PS-only constants.

TheMostDiligent (changes requested):

Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
hzqst: Done in 17ac310efd65dfa1848d95937d0c694a9d4fde10

Graphics/GraphicsEngineVulkan/src/PipelineStateVkImpl.cpp
hzqst: Done in 17ac310efd65dfa1848d95937d0c694a9d4fde10

Tests/DiligentCoreAPITest/src/SpecializationConstantsTest.cpp
hzqst: Done in 34bd28f4e67a844c7d4bb0bacd5a128a8f0cd568

Tests/DiligentCoreAPITest/src/SpecializationConstantsTest.cpp
hzqst: Done in 34bd28f4e67a844c7d4bb0bacd5a128a8f0cd568

TheMostDiligent:
> @TheMostDiligent why we have unrelated error with D3D12 tests even when I did nothing to D3D12 codepath?

This test sporadically fails: #319

Tests/DiligentCoreAPITest/src/ObjectCreationFailure/PSOCreationFailureTest.cpp
```diff
@@ -1759,4 +1759,122 @@ TEST_F(PSOCreationFailureTest, SpecConst_ErrorAtSecondElement)
 }
 
 
+// ---------------------------------------------------------------------------
+// Vulkan-specific negative tests for BuildSpecializationData (Name->SpecId matching).
+// These require GLSL shaders with layout(constant_id) declarations so that
+// the reflected specialization constants are available for matching.
+// ---------------------------------------------------------------------------
+
+TEST_F(PSOCreationFailureTest, SpecConst_NameNotFoundInShader)
```
TheMostDiligent: This should not trigger an error now

Tests/DiligentCoreAPITest/src/ObjectCreationFailure/PSOCreationFailureTest.cpp
```diff
@@ -1759,4 +1759,122 @@ TEST_F(PSOCreationFailureTest, SpecConst_ErrorAtSecondElement)
...
+    ShaderCI.SourceLanguage = SHADER_SOURCE_LANGUAGE_GLSL_VERBATIM;
+    ShaderCI.Desc           = {"SpecConst SizeMismatch CS", SHADER_TYPE_COMPUTE, true};
+    ShaderCI.Source         = CSSource_GLSL;
+
+    RefCntAutoPtr<IShader> pCS;
+    pDevice->CreateShader(ShaderCI, &pCS);
+    ASSERT_TRUE(pCS);
+
+    // Provide 8 bytes for a 4-byte float constant.
+    const double           WrongSizeData = 2.0;
```
TheMostDiligent: The size should be smaller now

Tests/DiligentCoreAPITest/src/ObjectCreationFailure/PSOCreationFailureTest.cpp
hzqst: Done in a3e881f13133214962d001a8918b048dc3de1279

PR #778, opened 2026-03-06
Add and Expose native handles and converter for NVidia Streamline

zackory137c-hash:
Diligent changes needed to expose certain handlers and converters to allow Streamline Integration for DLSS.

volk.c also need to be changed to allow for custom interception dll (sl.interposer.dll instead of vulkan-1.dll) but I can't get the submodule changes to commit.

### DIFF
--- a/Graphics/GraphicsEngineVulkan/include/TextureVkImpl.hpp
+++ b/Graphics/GraphicsEngineVulkan/include/TextureVkImpl.hpp
@@ -89,6 +89,11 @@ public:
         return StagingDataCPUAddress;
     }
 
+    VulkanUtilities::MemoryAllocation& GetMemoryAllocation()
+    {
+        return m_MemoryAllocation;
+    }
+
     void InvalidateStagingRange(VkDeviceSize Offset, VkDeviceSize Size);
 
     // For non-compressed color format buffer, the offset must be a multiple of the format's texel block size.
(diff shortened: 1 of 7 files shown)


TheMostDiligent:
We just added super resolution support, including DLSS.
https://github.com/DiligentGraphics/DiligentCore/tree/master/Graphics/SuperResolution/interface
These changes should not be needed now.

zackory137c-hash:
Are DLSS RR, FG, and the other toys in the DLSS / Streamline considered ?

TheMostDiligent:
Diligent exposes whatever [NVidia's DLSSS exposes](https://github.com/NVIDIA/DLSS)

Graphics/GraphicsEngineVulkan/include/TextureVkImpl.hpp
```diff
@@ -89,6 +89,11 @@ class TextureVkImpl final : public TextureBase<EngineVkImplTraits>
         return StagingDataCPUAddress;
     }
 
+    VulkanUtilities::MemoryAllocation& GetMemoryAllocation()
+    {
+        return m_MemoryAllocation;
+    }
```
TheMostDiligent, line 95: This should be const

TheMostDiligent (changes requested):

PR #791, opened 2026-05-21
Handle GPU timestamp overflow

Poohl:
Vulkan only guarantees 36 bits of timestamp, some Intel IGPUs (Core i7-14700HX) report
```
 timestampValidBits          = 36
 timestampPeriod                                 = 52.0833
```
resulting in an overflow after just one hour.

This commit implements proper handling for an overflow while measuring time (assuming you don't have two) and loggs a warning on overflow.

### DIFF
--- a/Graphics/GraphicsEngineVulkan/include/QueryManagerVk.hpp
+++ b/Graphics/GraphicsEngineVulkan/include/QueryManagerVk.hpp
@@ -72,6 +72,11 @@ public:
         return m_CounterFrequency;
     }
 
+    Uint32 GetTimestampValidBits() const
+    {
+        return m_TimestampValidBits;
+    }
+
     Uint32 ResetStaleQueries(const VulkanUtilities::LogicalDevice& LogicalDevice,
                              VulkanUtilities::CommandBuffer&       CmdBuff);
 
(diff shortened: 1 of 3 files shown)


TheMostDiligent:
This change does not seem to solve the problem. [The spec says](https://docs.vulkan.org/refpages/latest/refpages/source/VkQueueFamilyProperties.html):

> timestampValidBits is the unsigned integer count of meaningful bits in the timestamps written via [vkCmdWriteTimestamp2](https://docs.vulkan.org/refpages/latest/refpages/source/vkCmdWriteTimestamp2.html) or [vkCmdWriteTimestamp](https://docs.vulkan.org/refpages/latest/refpages/source/vkCmdWriteTimestamp.html). The valid range for the count is 36 to 64 bits, or a value of 0, indicating no support for timestamps. **Bits outside the valid range are guaranteed to be zeros.**


So ANDing bits with the mask should not really help much. In particular this line will not save from counter wrapping around and will return a very large value, which will incorrect. It should be clamped to zero instead.

`(EndCounter - StartCounter) & ((1ull << TimestampValidBits) - 1ull);`

Besides, the tests on CI fail and I expect this may happen because `TimestampValidBits` is 64.

Poohl:
I fixed the UB if `TimestampValidBits` is 64.

But the anding fixes the wrap around case, `EndCounter - StartCounter` will underflow (start counter is huge and end counter is small), causing all the invalid bits to become 1.

Small example with 16 bit uints and 8 valid bits:

```
end:      0x0003
start:    0x00FE
e-s:      0xFF05
mask:     0x00FF
out:      0x0005
             
```

PR #793, opened 2026-06-09
ShaderCreateInfo: Add SHADER_OPTIMIZATION_LEVEL

hzqst:
The `SHADER_OPTIMIZATION_LEVEL_DEFAULT` keeps what it was at 256017 with no regression.

### DIFF
--- a/Graphics/GraphicsEngine/include/DeviceObjectArchive.hpp
+++ b/Graphics/GraphicsEngine/include/DeviceObjectArchive.hpp
@@ -128,7 +128,7 @@ public:
     };
 
     static constexpr Uint32 HeaderMagicNumber = 0xDE00000A;
-    static constexpr Uint32 ArchiveVersion    = 9;
+    static constexpr Uint32 ArchiveVersion    = 10;
 
     struct ArchiveHeader
     {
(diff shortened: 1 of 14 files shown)


TheMostDiligent:
This new value needs to be serialized in [SerializeCI](https://github.com/DiligentGraphics/DiligentCore/blob/ed450f2e321da16e02efee80237d0f22227e342e/Graphics/GraphicsEngine/src/PSOSerializer.cpp#L508)

hzqst:
> This new value needs to be serialized in [SerializeCI](https://github.com/DiligentGraphics/DiligentCore/blob/ed450f2e321da16e02efee80237d0f22227e342e/Graphics/GraphicsEngine/src/PSOSerializer.cpp#L508)

image
like this ?

TheMostDiligent:
> > This new value needs to be serialized in [SerializeCI](https://github.com/DiligentGraphics/DiligentCore/blob/ed450f2e321da16e02efee80237d0f22227e342e/Graphics/GraphicsEngine/src/PSOSerializer.cpp#L508)
> 
> image like this ?

Yes, missed it. Thanks

